PR #20, opened 2023-06-26
16 porting to macos

### LINKED ISSUE #16: Port Khiops to macOS
- Review CMakeFiles to remove `__UNIX__`  pragmas.
- modify Norm module to build on macOS

bruno-at-orange:
- The build on macOS is OK except for gtest where regex.h is missing (https://github.com/KhiopsML/khiops/actions/runs/5376228304/jobs/9753105311).
- For the compiler options, the only problem was the hardening flags mandatory for debian. I have excluded them for macOS.
- There are a lot of warnings while building.
- The build was not parallelized without the --parallel option on the build preset (it should be parallel by default but it is not the case on macos)

I propose to merge this branch and to open 2 issues (one for the warnings, one for google test). It will allow us to work on other features (conda build...) and to invite Alexis on github.


folmos-at-orange:
~~Hello, before merging you should rebase to `dev`~~

My mistake :D. However you may want to squash your commits into a single one, be careful not to squash the merge commit.

 I'm trying to working a fix for macos, let me see if I can do something before the merge.

src/Learning/KNITransfer/CMakeLists.txt
```diff
@@ -4,4 +4,4 @@ add_executable(KNITransfer ${cppfiles})
 set_khiops_options(KNITransfer)
 
 target_link_libraries(KNITransfer PUBLIC KhiopsNativeInterface KWLearningProblem DTForest KMDRRuleLibrary
-                                         $<$<BOOL:${MPI}>:PLMPI> PLParallelTask)
```
folmos-at-orange, line 7: MPI is not necessary for KNITransfer ?

src/Norm/base/ALString.h
```diff
@@ -307,7 +307,7 @@ inline ALString::ALString()
 {
 	Init();
 }
-#if not defined __UNIX__ or defined __C11__
+#if defined __C11__
```
folmos-at-orange, line 310: is `__C11__` defined when C++14 is used ?

folmos-at-orange (approved):

src/Norm/base/ALString.h
bruno-at-orange: You're Hawkeye ! This a bug  `__C11__`  was defined in the CMakeLists that's no longer the case. I will change it with `__cplusplus`

src/Learning/KNITransfer/CMakeLists.txt
bruno-at-orange: Yes, I forgot this one, I have remove all these  `$<$<BOOL:${MPI}>:PLMPI>` to add only one in `PLParallelTask`. So `PLMPI` will be add transitively by any targets adding `PLParallelTask`.

src/Norm/base/ALString.h
folmos-at-orange: Ok !

bruno-at-orange: Actualy, this is OK, we have those lines in portability.h
```cpp
// Utilisation de c++11
#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
#define __C11__
#endif
```

PR #21, opened 2023-06-26
Fix namespace problem for Regex.h

folmos-at-orange:
src/Norm/base/Regex.h collided with that of the standard C++ library making fail the compilation of unit tests in macOS.


bruno-at-orange (approved):
LGFM

PR #26, opened 2023-06-28
3 use targets in cmake to speed up the build process

### LINKED ISSUE #3: Use targets in cmake to speed up the build process
We can specify the targets to be build:
```bash
cmake --build . --target target1 target2
```
It will speed up the build process (especially for unit test)

bruno-at-orange:
- Use target in cmake build
- Remove several options in the main CMakeLists:
    - FULL, LEARNING and PARALLEL: not needed anymore (use targets)
    - MODL_SO: binary MODL_dll is always built
    - FEDORA: the fedora distro is discovered automatically and a fatal error is triggered if MPI_SUFFIX is not set by the mpi module
- Change the name of some test targets: all in lowercase
- CMakePresets:
    - add `"jobs":0` to enable parallel build
    -  simplification of the naming: remove the prefix "build_" in the name of the build preset


.github/workflows/pack-debian.yml
```diff
@@ -26,14 +22,13 @@ jobs:
...
-        echo "KHIOPS_VERSION=$KHIOPS_VERSION" >> "$GITHUB_ENV"
+       
 
     - name: Configure CMake
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DMPI=ON -DTESTING=OFF -DFULL=OFF -DBUILD_JARS=ON
+      run: cmake -B ${{github.workspace}}/build -DMPI=ON -DTESTING=OFF -DBUILD_JARS=ON
 
     - name: Build
-      run: time cmake --build ${{github.workspace}}/build --parallel
+      run: time cmake --build ${{github.workspace}}/build --parallel --target MODL MODL_Coclustering KhiopsNativeInterface
```
folmos-at-orange: No need to use `time`. Github itselft times it

.github/workflows/pack-rpm.yml
```diff
@@ -32,10 +32,11 @@ jobs:
       run:  |
         source /etc/profile.d/modules.sh
         module load mpi/mpich-x86_64
-        cmake -B build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DMPI=ON -DTESTING=OFF -DFULL=OFF -DBUILD_JARS=ON -DFEDORA=ON
+        echo "MPI_SUFFIX=$MPI_SUFFIX" >> "$GITHUB_ENV"
+        cmake -B build -DTESTING=OFF -DBUILD_JARS=ON -DFEDORA=ON
 
     - name: Build
-      run: time cmake --build build --parallel
+      run: cmake --build build --parallel --target MODL${{ env.MPI_SUFFIX }} MODL_Coclustering KhiopsNativeInterface
```
folmos-at-orange, line 36: I think you are building two times. The configure step does not use a preset nor uses `-C`

.github/workflows/run-unit-tests.yml
```diff
@@ -30,15 +30,20 @@ jobs:
...
 
-      - name: Run Unit Tests for the Parallel Module
+      - name: Run Unit Tests for the Parallel MPI Module in serial
         if: success() || failure()
-        run: build/${{ matrix.build-setup.cmake-preset }}/bin/Parallel_test --gtest_output="xml:${{github.workspace}}/reports/report-parallel.xml"
+        run: build/${{ matrix.build-setup.cmake-preset }}/bin/parallel_test --gtest_output="xml:${{github.workspace}}/reports/report-parallel.xml"
+
+      - name: Run Unit Tests for the Parallel MPI Module in parallel
+        if: success() || failure()
+        run: mpiexec  -n 4 build/${{ matrix.build-setup.cmake-preset }}/bin/parallel_mpi_test --gtest_output="xml:${{github.workspace}}/reports/report-parallel-mpi.xml"
```
folmos-at-orange: Nitpick: Remove extra space here.

CMakeLists.txt
```diff
@@ -276,18 +288,10 @@ function(display_variables)
   endforeach()
 endfunction()
 
-# Always build Norm
+# Build Norm Parallel and Learning modules
```
folmos-at-orange: "Build" is confusing as it we are only defining targets. Maybe "Add targets for Norm Parallel and Learning modules" ?

README.md
```diff
@@ -25,13 +25,13 @@ cd build
 cpack -G DEB
 ```
 
-On Fedorra-like distro, whe have to set `FEDORA` to `ON` to configure the cmake project. Moreover, the environment module mpi must be loaded before configure, build and package processes:
+On Fedorra-like distro, the environment module mpi must be loaded before configure, build and package processes:
```
folmos-at-orange: Fedo**rr**a -> Fedora

packaging/install.cmake
```diff
@@ -125,7 +125,7 @@ else(NOT FEDORA)
     DESTINATION usr/bin
     COMPONENT KHIOPS_CORE)
 
-endif(NOT FEDORA)
+endif(NOT IS_FEDORA_LIKE)
```
folmos-at-orange: We should discus whether use this CMake feature of "commenting" the `endif`. I find it confusing, mostly when there are `else` and or `elseif` expressions.

.github/workflows/pack-rpm.yml
folmos-at-orange: Also in the configure step shouldn't `CMAKE_BUILD_TYPE=Release` . I would put it even if it is the default value.

folmos-at-orange (changes requested):

folmos-at-orange:
Another point: The macos parallel unit-test are failing because of java (???). Maybe it is easily fixable.

Otherwise a very nice simplification PR :)

.github/workflows/pack-rpm.yml
bruno-at-orange: nope `cmake -B build` configure and `cmake --build build` build....
However `CMAKE_BUILD_TYPE=Release` is missing, I think the default value is 'Debug`

bruno-at-orange:
For the macos unit-test, I don't think it is because of java: java is not used !

folmos-at-orange:
> For the macos unit-test, I don't think it is because of java: java is not used !

The failed test says `fatal error : Java GUI : Java GUI is not available on MacOS`

See: https://github.com/KhiopsML/khiops/actions/runs/5402237529/jobs/9813233296?pr=26#step:6:55

bruno-at-orange:
> > For the macos unit-test, I don't think it is because of java: java is not used !
> 
> The failed test says `fatal error : Java GUI : Java GUI is not available on MacOS`
> 
> See: https://github.com/KhiopsML/khiops/actions/runs/5402237529/jobs/9813233296?pr=26#step:6:55

It is weird because on runners (windows and linux), GUI is not available and tests passed. There is an issue in macOS porting.

.github/workflows/pack-rpm.yml
folmos-at-orange: Ok !

folmos-at-orange:
> > > For the macos unit-test, I don't think it is because of java: java is not used !
> > 
> > 
> > The failed test says `fatal error : Java GUI : Java GUI is not available on MacOS`
> > See: https://github.com/KhiopsML/khiops/actions/runs/5402237529/jobs/9813233296?pr=26#step:6:55
> 
> It is weird because on runners (windows and linux), GUI is not available and tests passed. There is an issue in macOS porting.

It is really weird. Since it works in serial maybe it is a slave process that tries to instantiate an UIObject or something like that. I will make an issue of it.

folmos-at-orange:
Last issue: RPM packaging is broken because of the missing pdfs. Maybe we should fix that before merging ? Since you changed the build process it would be good to verify that the packaging processes work as well.

bruno-at-orange:
> Last issue: RPM packaging is broken because of the missing pdfs. Maybe we should fix that before merging ? Since you changed the build process it would be good to verify that the packaging processes work as well.

OK I will do it

folmos-at-orange (approved):

PR #35, opened 2023-06-30
29 unit tests in parallel fail in macos

### LINKED ISSUE #29: Unit tests in parallel fail in MacOs
See [here](https://github.com/KhiopsML/khiops/pull/26#issuecomment-1611663131) for context

### LINKED ISSUE #32: the "jobs" option in the build preset fails on windows
On Visual Studio 2022, setting jobs to 0 is an issue

bruno-at-orange:
In addition to the fix of the unit test I handled the [issue 32](https://github.com/KhiopsML/khiops/issues/32#issue-1781096435). A problem with the "jobs" option in the CMakePreset.

For this one the only way is to suppress this flag, there is no solution to add a max number of parallel level, it has to be an integer (there is an [open issue on cmake](https://gitlab.kitware.com/cmake/cmake/-/issues/23798)). In consequence, we have to add the `--parallel` flag to the `cmake --build` on the actions. The problematic issue is that on ours IDE, the build are in  serial. The solution to this is to add a `CMakeUserPresets` with the jobs  option.

### DIFF
--- a/.github/workflows/run-unit-tests.yml
+++ b/.github/workflows/run-unit-tests.yml
@@ -37,13 +37,17 @@ jobs:
         if: success() || failure()
         run: build/${{ matrix.build-setup.cmake-preset }}/bin/norm_test --gtest_output="xml:${{github.workspace}}/reports/report-norm.xml"
 
-      - name: Run Unit Tests for the Parallel MPI Module in serial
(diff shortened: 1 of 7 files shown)


folmos-at-orange (approved):
LGTM

PR #38, opened 2023-07-03
Build macos conda package

### LINKED ISSUE #22: Build macOS conda package

bruno-at-orange:
Conda package is built but never tested.
I added a workflow for macOS packaging: It is a simple zip. The purpose is to make available the binaries for Khiops developers for tests. We can look further to `CPack` generators for macOS as `productbuild`, `FreeBSD` or `DragNDrop`. I'm not sure it's worth the effort: we will not distribute macOS installers.

### DIFF
--- a/.github/workflows/pack-macos.yml
+++ b/.github/workflows/pack-macos.yml
@@ -1,0 +1,32 @@
+name: Create macOS package
+  
+on:
+  workflow_dispatch:
+
+env:
+  PRESET_NAME: macos-clang-release
+
+jobs:
+  package-zip:
+    runs-on: macos-latest
+    
+    steps:
+    - name: Check out sources
+      uses: actions/checkout@v3
+
+    - name: Build Khiops executables
+      uses: ./.github/actions/build-khiops
+      with:
+        preset-name: ${{env.PRESET_NAME}}
+        targets: "MODL MODL_Coclustering KhiopsNativeInterface"
+        override-flags: -DBUILD_JARS=OFF -DTESTING=OFF
+     
+    - name: Build package with CPack
+      run: cd build/${{env.PRESET_NAME}} && cpack -G ZIP
+    
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v3.1.2
+      with:
+        name: macos
+        path: build/${{env.PRESET_NAME}}/packages/*.zip
+        if-no-files-found: error
--- a/packaging/conda/build.sh
+++ b/packaging/conda/build.sh
@@ -1,3 +1,16 @@
 # !/bin/bash
-cmake --preset linux-gcc-release -DBUILD_JARS=OFF -DTESTING=OFF
-cmake --build --preset linux-gcc-release --parallel --target MODL --target MODL_Coclustering
+
+# On macOS, we have to build with the compiler outside conda. With teh conda's clang the following error occurs:
+# ld: unsupported tapi file type '!tapi-tbd' in YAML file '/Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/lib/libSystem.tbd' for architecture x86_64
+if [ "$(uname)" == "Darwin" ]
+then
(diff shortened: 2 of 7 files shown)


.github/workflows/pack-macos.yml
```diff
@@ -0,0 +1,32 @@
...
+
+env:
+    PRESET_NAME: macos-clang-release
+
+jobs:
+  package-zip:
+    runs-on: macos-latest
+    
+    steps:
+    - uses: actions/checkout@v3
```
folmos-at-orange: For clarity add a title (eg "Check out sources")

.github/workflows/pack-macos.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: Build Khiops test executables
```
folmos-at-orange: "Build Khiops executables"  (not **test**)

.github/workflows/pack-macos.yml
```diff
@@ -0,0 +1,32 @@
...
+      uses: ./.github/actions/build-khiops
+      with:
+          preset-name: ${{env.PRESET_NAME}}
+          targets: "MODL MODL_Coclustering KhiopsNativeInterface"
+          override-flags: -DBUILD_JARS=OFF -DTESTING=OFF
+
```
folmos-at-orange, line 23: Remove white space (maybe we should add a YAML formatter?)

.github/workflows/pack-macos.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Build Khiops test executables
+      uses: ./.github/actions/build-khiops
+      with:
+          preset-name: ${{env.PRESET_NAME}}
+          targets: "MODL MODL_Coclustering KhiopsNativeInterface"
+          override-flags: -DBUILD_JARS=OFF -DTESTING=OFF
+     
+  
+    - name: Pack
```
folmos-at-orange: Build package with CPack

packaging/conda/build.sh
```diff
@@ -1,3 +1,16 @@
 # !/bin/bash
-cmake --preset linux-gcc-release -DBUILD_JARS=OFF -DTESTING=OFF
-cmake --build --preset linux-gcc-release --parallel --target MODL --target MODL_Coclustering
+
+# On macOS, we have to build with the compiler outside conda. With teh conda's clang the following error occurs:
+# ld: unsupported tapi file type '!tapi-tbd' in YAML file '/Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/lib/libSystem.tbd' for architecture x86_64
```
folmos-at-orange, line 4: We'll not be sure that the fix works until we test its installation. I have doubts that compiling with the system's compiler will work because conda is made to work in its environment (but maybe I'm wrong)

packaging/conda/install-khiops.sh
```diff
@@ -3,7 +3,16 @@
 # Echo all output
 set -x
 
+
```
folmos-at-orange: Remove extra whitespace

folmos-at-orange (changes requested):

.github/workflows/pack-macos.yml
bruno-at-orange: OK added the same titles in pack-debian and pack-rpm

.github/workflows/pack-macos.yml
bruno-at-orange: yes !
all that can be automatically formatted should be formatted...

packaging/conda/build.sh
bruno-at-orange: Yes indeed this is not the good way to do it. We have to test this package on an OS different than the one used in the build process e.g. macOS 12 Monterey.

folmos-at-orange: If this doesn't work we should give a shot to the fix mentioned here : https://computing.docs.ligo.org/conda/compiling/#sdk

folmos-at-orange (approved):
LGTM (If the package doesn't work we'll open another issue)

PR #60, opened 2023-08-02
Fix in KWFileSorter::SplitDatabase

bruno-at-orange:
Assertion triggered in `LearningTest\TestKhiops\CrashTests\MaxLineLength_sort `
The sample may be larger the line number.


marcboulle (approved):
LGTM (deja pris en compte dans V11)

PR #67, opened 2023-08-07
Set default CMake generators per platform

### LINKED ISSUE #58: Optimize performance for release binaries with Visual C++ 2022
Les performances se sont fortement dégradées depuis le passage à Visual C++ 2022, d'un facteur 3 à 4.
Le paramétrage des options de compilation via CMake doit être optimisé.

Exemples sur ma machine avec la version V10.4.6i:
- LearningTest\TestKhiops\Standard\Adult
  - Data preparation time: 0:00:00.30 -> 0:00:01.05
  - SNB train time: 0:00:00.68 -> 0:00:02.26
  - Model deployment time: 0:00:00.09 ->0:00:00.38
- LearningTest\TestKhiops\Classification\ForestType
  - Data preparation time: 0:00:04.45 -> 0:00:17.79
  - SNB train time: 0:01:40.30 -> 0:05:28.17
  - Model deployment time: 0:00:00.18 -> 0:00:00.95

folmos-at-orange:
We settle for `Ninja` in Windows and `Unix Makefiles` in Linux and MacOs.

### DIFF
--- a/.github/actions/build-khiops/action.yml
+++ b/.github/actions/build-khiops/action.yml
@@ -15,6 +15,8 @@ runs:
   steps:
     - name: Setup MPI
       uses: mpi4py/setup-mpi@v1
+    - name: Setup Ninja
+      uses: ashutoshvarma/setup-ninja@v1.1
     - name: Load Visual C++ Environment Variables (Windows only)
       if: ${{ runner.os == 'Windows' }}
       shell: cmd
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -9,32 +9,28 @@ jobs:
       fail-fast: false
       matrix:
         os: [rocky8, rocky9, ubuntu18.04, ubuntu20.04, ubuntu22.04]
-
-    # Allows to write on the package registry
     permissions:
-      packages: write
+      packages: write  # Allows writing in the container registry
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v2
-      - name: Login to Github Packages
+      - name: Login to Github Container Registry
         uses: docker/login-action@v2
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - name: Build image and push to GitHub Container Registry
+      - name: Build the image and push it to the registry
         uses: docker/build-push-action@v2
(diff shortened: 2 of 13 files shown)


bruno-at-orange (changes requested):
On mac and Linux, I think it is better to let the default generator.
Why do we have to use makefile rather than Ninja ?

folmos-at-orange:
> On mac and Linux, I think it is better to let the default generator.
> Why do we have to use makefile rather than Ninja ?

The default generator in Linux and Mac are Makefiles. I think it is better to control explicitly the generator to have better reproducibility. The "default" generator may depend on the user's machine as CMake does many things automatically.

bruno-at-orange:
I don't like the idea to force developers to use a particular generator on their own environment. On production it is different, but we can specify the generator outside the preset (with the `-G` flag).
If you stuck on the idea to set it on the preset,  why did you choose Makefile ? Ninja is known to be faster and it builds natively in parallel (this is not the case for Makefile, whe have to add `--parallel`)...

folmos-at-orange:
After discussion we'll set the default generator for all platforms as `Ninja`

folmos-at-orange:
In addition to setting ninja-build as default for all presets:
- I installed ninja in all packaging containers
- I use presets in the linux build (for reproducibility)
  - So I installed a more recent version of cmake in the containers (from kitware's site)
- I renamed the containers to `khiopsml/khiops/khiopsdev` because the previous naming left them in the organizations repository and I couldn't overwrite them. I chose `khiopsdev` because people may confuse it with a docker **with** khiops inside. Additionally I set their visibility to "Internal".

.github/actions/build-khiops/action.yml
```diff
@@ -15,6 +15,8 @@ runs:
   steps:
     - name: Setup MPI
       uses: mpi4py/setup-mpi@v1
+    - name: Setup Ninja
```
bruno-at-orange, line 18: I think you should add `if: ${{ runner.os != 'Linux' }}` because ninja is already on the docker images for Linux

bruno-at-orange (approved):
LGTM 
perhaps a minor change in .github/actions/build-khiops/action.yml

.github/actions/build-khiops/action.yml
folmos-at-orange: Yes, but when we run the tests we do in on Github's VM, and there is no Ninja

bruno-at-orange: OK...

PR #121, opened 2024-01-11
120 jar files are not present in the ubuntu debian khiops desktop packages

### LINKED ISSUE #120: JAR files are not present in the Ubuntu / Debian `khiops` desktop packages
#### Description
When following the official instructions for installing Khiops Desktop on Ubuntu (22.04) - see https://khiops.org/setup/nocode/#__tabbed_1_2 - the JAR files `khiops.jar` and `norm.jar` are not installed in `/usr/share/khiops`, as expected by the `khiops-env` script. Hence, the Java-based Khiops GUI fails to start.

#### Questions/Ideas
- The issue is that in the CMake install script that `cpack` uses there is no `install` directive to include these JAR files in the `KHIOPS` component. Hence, these JAR files are not included in the `khiops_*.*.deb` packages, but are included in the `khiops-unspecified_*.*.deb` packages instead which are not (supposed to be) released.
- The solution is to include these two JAR files in an `install` directive for the `KHIOPS` component.
  - Moreover, we can also exclude the "Unspecified" component from being used for generating the `khiops-unspecified_*.*.deb` package.
- Note: This is expected to also impact RPM packages, in exactly the same way, as the issue seems to be not Debian-related, but CPack-related.
- Note: We need to see how / whether this can be tested on the CI.
### Context
- Khiops version: Khiops 10.2.0
- Log file (use `khiops -e log.txt`): its contents are
   ```
   fatal error : Java Library : Please check that Java version is at least Java 7, that the X server is avai 
   lable under current environment and check access to the norm.jar library in the Java classpath
   ```
- Scenario file (use `khiops -o scenario._kh`): not applicable (calling `khiops` fails to launch the GUI and exits)
- OS description (use `khiops -s`): this command exits with `Unexpected -s flag`.

popescu-v:
closes #120

### DIFF
--- a/.github/workflows/pack-debian.yml
+++ b/.github/workflows/pack-debian.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         os: [debian10, debian11, ubuntu22.04, ubuntu20.04]
     container:
-      image: ghcr.io/${{ github.repository }}/khiopsdev-${{ matrix.os }}:latest
+      image: ghcr.io/khiopsml/khiops/khiopsdev-${{ matrix.os }}:latest
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
@@ -106,7 +106,7 @@ jobs:
       matrix:
         os: [debian10, debian11, ubuntu22.04, ubuntu20.04]
     container:
-      image: ghcr.io/${{ github.repository }}/khiopsdev-${{ matrix.os }}:latest
+      image: ghcr.io/khiopsml/khiops/khiopsdev-${{ matrix.os }}:latest
     steps:
       - name: Put OS info on the environment
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/pack-debian.yml
folmos-at-orange, line 1: Can't we make the same trick for  as in `build-linux-pack-containers.yml` for `GH_REPOSITORY` ?

folmos-at-orange (changes requested):
I'd squash all the commits related to the packaging.

.github/workflows/pack-debian.yml
popescu-v: I don't see how, because in that other case, I had added a preliminary stage doing this. Whereas here, we would need this variable from the outset (before any job stage is executed).

popescu-v:
> I'd squash all the commits related to the packaging.

I would too, but I wanted to make sure we agree on this before squashing.

.github/workflows/pack-debian.yml
folmos-at-orange: Ok, let's leave it like that for the moment.

folmos-at-orange (approved):

PR #158, opened 2024-02-19
Fix JAVA_HOME setting in Debian post-install script

### LINKED ISSUE #157: Khiops Desktop does not launch on Ubuntu
#### Description
On Ubuntu systems, after installing the latest release `khiops` package, the `khiops` command fails to launch the desktop GUI and aborts. The log gives:
```
$ khiops -e error.log
$ cat error.log
error : Java : Unable to load libjvm.so (libjvm.so: cannot open shared object file: No such file or directory), install java JRE or check the LD_LIBRARY_PATH variable
error : Unable to load GUI, use -b and -i flags to launch khiops in batch mode (khiops -h for help)
```
#### Questions/Ideas
- The issue is that the `JAVA_HOME` environment variable is not set in the `khiops-env` script by the Debian post-install script.
  - Indeed, the aforementioned post-install script sets `JAVA_HOME` in the `khiops` script instead. The fix consists in setting this variable in the `khiops-env` script.
- The bug is not visible if the `JAVA_HOME` environment variable is correctly set in the host shell by other means, prior to launching the `khiops` command.

### Context
- Khiops version: 10.2.0
- Log file (use `khiops -e log.txt`)
  ```sh
  error : Java : Unable to load libjvm.so (libjvm.so: cannot open shared object file: No such file or directory), install java JRE or check the LD_LIBRARY_PATH variable
  error : Unable to load GUI, use -b and -i flags to launch khiops in batch mode (khiops -h for help)
  ```
- Scenario file (use `khiops -o scenario._kh`): not applicable; just launching the basic `khiops` command triggers the bug
- OS description (use `khiops -s`): Ubuntu 22.04 (LTS), but this should apply to all Debian-based OSes (as the `postinst` script is the same)

popescu-v:
closes #157

### DIFF
--- a/packaging/linux/debian/khiops/postinst
+++ b/packaging/linux/debian/khiops/postinst
@@ -4,11 +4,11 @@ set -e
 # setting JAVA_HOME
 if [ -e /usr/bin/java ]; then
     JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:bin/java::")
-    sed -i -e "s|# JAVA_HOME automatic setting|[[ -z \"\$JAVA_HOME\" ]] \&\& JAVA_HOME=$JAVA_HOME|g" /usr/bin/khiops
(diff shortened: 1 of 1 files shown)


bruno-at-orange (approved):
LGTM

PR #194, opened 2024-03-18
183 transform the basetest binary into unit tests

### LINKED ISSUE #183: Transform the basetest binary into unit tests
The basetest binary is used to test the base library. At this time basetest is launched manually. We have to transform it into unit tests.

bruno-at-orange:
- Add all methods of basetest in unit tests (except 2 with GUI)
- Add 2 methods of Learning in unit tests (one is buggy with an assertion and is commented)
- mv unit test directories to test/UnitTests
- set TESTING option default to ON in the preset (and add -DTESTING=OFF when it is needed in the actions)
- Improve the testing action in order to handle crashes in the unit test macro (assertion). results files are available for download, when test fails


src/Norm/base/ALString.cpp
marcboulle, line 1: ligne 8: corriger le caractere accentué qui ne passe pas

marcboulle (approved):
LGTM

test/Norm/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 file(GLOB cppfiles *.cpp)
 add_executable(norm_test ${cppfiles})
 set_khiops_options(norm_test)
-target_link_libraries(norm_test GTest::gtest_main base)
+target_link_libraries(norm_test GTest::gtest_main base testutils)
```
folmos-at-orange, line 4: Should this be done only if `TESTING=ON`?

test/Utils/TestServices.h
```diff
@@ -59,8 +64,8 @@
 	}
 
 // Comparaison de 2 fichiers ligne par ligne
-// Les fichiers peuvent differer pour les lignes qui commencent par 'SYS'
-// Il y a egalement une tolerance pour le caractere separateur dans les chemins '\' qui est remplace par '/'
+// Si deux lignes contiennet le token 'SYS', elles sont comparees uniquement jusqu'a ce token: tout ce qui est derriere
```
folmos-at-orange: contiennet -> contiennent

.github/workflows/run-unit-tests.yml
```diff
@@ -74,6 +74,31 @@ jobs:
...
+            test/KNITest/results/ \
+            test/Norm/results/ \
+            test/Learning/results/ \
+            test/Norm/results/ \
+            test/Parallel/results/ \
+            test/Parallel-mpi/results/); then
+            echo "::error::At least one test binary has failed with an assertion (assert, check, require, ensure..). Check result files available for download as artifacts."
+          fi
+      - name: Upload test results
+        if: failure()
```
folmos-at-orange, line 90: I think it doesn't cost much to always provide the files. It may be useful if there is some kind of malfunction and the `grep` doesn't catch it.

CMakePresets.json
```diff
@@ -13,7 +13,7 @@
       "generator": "Ninja",
       "cacheVariables": {
         "MPI": "ON",
-        "TESTING": "OFF",
+        "TESTING": "ON",
```
folmos-at-orange, line 16: This default setting should be also replicated in the `CMakeLists.txt`

folmos-at-orange (approved):
LGTM

I think it could be squashed when merging.

test/Norm/CMakeLists.txt
bruno-at-orange: This CMakeLists is added only if TESTING=ON

.github/workflows/run-unit-tests.yml
bruno-at-orange: hmm, I disagree, if any test fails (or the test binary crashes), the results will be uploaded. 
The grep is here only to inform the user that there is an assertion
On the other hand it all tests pass why do you want to see the results, they are the same as results.ref

CMakePresets.json
bruno-at-orange: OK

.github/workflows/run-unit-tests.yml
folmos-at-orange: Ok

test/Norm/CMakeLists.txt
folmos-at-orange: Ok

PR #198, opened 2024-03-20
Improve khiops-env and packages versioning

### LINKED ISSUE #99: khiops-env linux/mac script takes a long time to find libjvm
#### Description
The `khiops-env` script may take a long time to find `libjvm.so` in some environments. For example in a default conda environment with openjdk installed khiops-env takes 1 minute in my machine. This is because `JAVA_HOME` is set to the root of the conda installation and thus the `find` command explores *all* its sub-directories.

This is critical for the `khiops-python` + Khiops system wide installation as it relies on `khiops-env` to obtain the environment but Java will not be required in newer versions as the `batch` option is deprecated and it will be removed with Khiops 11.

#### Questions/Ideas
- Two options
  - Improve the search heuristic: search only in selected sub-directories of `JAVA_HOME`
  - Add an option to skip the `libjvm.so` lookup.

### Context
- Khiops 10.x
- Linux/Mac

bruno-at-orange:
- In khiops-env, the jvm.so search is enabled only if the khiops packages is installed. 
- The new versioning scheme MAJOR.MINOR.PATCH{-PRE_RELEASE.PR_INCREMENT} is now handled in the Linux packages:
  - in the packages names e.g. khiops_10.5.0-a1-1_amd64.deb or kni-10.5.0_a1-1.x86_64.rpm
  - in the so files e.g. libKhiopsNativeInterface.so.10.5.0-a1


folmos-at-orange (approved):
LGTM

PR #206, opened 2024-03-22
Fix wrong errno in remote fread

bruno-at-orange:
In the class `PLMPISystemFileDriverRemote`, we use a new class attribute `nLocalErrno` to store errno code. We can not relies on the global errno for file access because `MPI_Recv` modifies errno (sets it to 11, Resource Temporarily Unavailable) even if all is OK.


folmos-at-orange (approved):
LGTM

PR #210, opened 2024-03-25
Fix package renaming in CI

folmos-at-orange:
commit 11c4c6cd5415a1ad7f59b84eb2ea1c83e3c25737 (HEAD -> fix-renaming-nsis, origin/fix-renaming-nsis)
Author: Felipe Olmos <92923444+folmos-at-orange@users.noreply.github.com>
Date:   Tue Mar 5 19:04:55 2024 +0100

    Allow CI failure for unsupported Debian versions

    We use `continue-on-error` for certain steps of the testing packages for
    Debian 11 and 12. This way the CI is green even though in these OS the
    packages fail on basic tests.

commit bfa6aa362d684472a48290c431f87aa99b287fc2 (tag: 10.2.1)
Author: Felipe Olmos <92923444+folmos-at-orange@users.noreply.github.com>
Date:   Mon Mar 25 10:25:32 2024 +0100

    Fix package renaming in CI

    The release part of the packaging workflows contain a renaming step to
    make sure it coincides with the tag. This patch makes sure that this is
    only made when the source version is different from the tag.

### DIFF
--- a/.github/workflows/pack-debian.yml
+++ b/.github/workflows/pack-debian.yml
@@ -65,7 +65,6 @@ jobs:
   test:
     needs: build
     runs-on: ubuntu-latest
-    continue-on-error: true
     defaults:
       run:
         shell: bash
@@ -94,6 +93,7 @@ jobs:
           dpkg -i ./artifacts/khiops-core* || true
           apt-get -f install -y
       - name: Check Khiops core installation
+        continue-on-error: ${{ matrix.os == 'debian:11' }}
         run: |
           khiops-env --env
           khiops -v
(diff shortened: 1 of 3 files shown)


bruno-at-orange (approved):
LGTM

bruno-at-orange (approved):
LGTM
I am not very satisfied with the  `continue-on-error` for the debian but I don't have better ideas

folmos-at-orange:
> LGTM I am not very satisfied with the `continue-on-error` for the debian but I don't have better ideas

Me neither :(. I tried to constrained as much as possible, before it was the whole job that was with `continue-on-error`. And it is ugly enough so we won't forget to remove once the build for Debian works.

We could emit a warning on failure of these particular steps (`echo ::warning`) but it complexifies things a bit too much I think.

PR #214, opened 2024-03-27
Reincorporate conda packaging from khiops-python

folmos-at-orange:

PR #226, opened 2024-04-08
Update conda packaging in dev

folmos-at-orange:
Same as #214  but for `dev`


.gitignore
```diff
@@ -20,7 +20,11 @@ CMakeFiles/
 # Test exceptions
 test/UnitTests/*/results/
 test/LearningTest/*/*/*/results/
+test/*/results/
+test/LearningTest/*/*/*/results.ref/time.log
+results/
```
bruno-at-orange: I think line 23 and 25 are redundant with 21 and 22. We can choose to set only line 25 over 21 22 and 23 (to discuss)
line 24 is wrong, we have to keep time.log in results.ref

bruno-at-orange (approved):
LGTM
just modify the .gitignore file

.gitignore
folmos-at-orange: Ok

folmos-at-orange: Just not to have a range too large how about `test/**/results/` ?

PR #228, opened 2024-04-09
203 switch to openmpi for linux packages

### LINKED ISSUE #203: Switch to OpenMPI for Linux packages
OpenMpi seems better than MPICH for multiple reasons (well documented, better packaging etc...). We have to study the best way to switch from MPICH to OpenMPI. Ideally, we should be able to deliver a package for mpich and a package for intelmpi. 
This issue is  related to #164

bruno-at-orange:


bruno-at-orange:
wrong branch

Issue #164, opened 2024-02-21
MPICH issue on debian 11


bruno-at-orange:
#### Description
On debian 11, khiops raise an mpi error at launch:

#### Questions/Ideas
- It is a packaging issue of mpich (hwloc is missing)
- We can wait a new release of mpich or we can switch to another flavor of mpi: `Open MPI` or `Intel MPI`

### Context
- Khiops version 10.2.0
- Debian 11
- stderr:

```
[proxy:0:0@8f731690f735] HYDT_topo_bind (tools/topo/topo.c:89): no topology library available
[proxy:0:0@8f731690f735] HYDU_create_process (utils/launch/launch.c:66): bind process failed
[proxy:0:0@8f731690f735] launch_procs (pm/pmiserv/pmip_cb.c:730): create process returned error
[proxy:0:0@8f731690f735] HYD_pmcd_pmip_control_cmd_cb (pm/pmiserv/pmip_cb.c:904): launch_procs returned error
[proxy:0:0@8f731690f735] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[proxy:0:0@8f731690f735] main (pm/pmiserv/pmip.c:169): demux engine error waiting for event
[proxy:0:0@8f731690f735] HYDT_topo_bind (tools/topo/topo.c:89): no topology library available
[proxy:0:0@8f731690f735] HYDU_create_process (utils/launch/launch.c:66): bind process failed
[proxy:0:0@8f731690f735] launch_procs (pm/pmiserv/pmip_cb.c:730): create process returned error
[proxy:0:0@8f731690f735] HYD_pmcd_pmip_control_cmd_cb (pm/pmiserv/pmip_cb.c:904): launch_procs returned error
[proxy:0:0@8f731690f735] HYDT_dmxu_poll_wait_for_event (tools/demux/demux_poll.c:76): callback returned error status
[proxy:0:0@8f731690f735] main (pm/pmiserv/pmip.c:169): demux engine error waiting for event

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 1301 RUNNING AT 8f731690f735
=   EXIT CODE: 9
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Killed (signal 9)
[...]

PR #229, opened 2024-04-09
203 switch to openmpi for linux packages

### LINKED ISSUE #23: Use CTest to run unit tests with mpi
We can setup mpi run in the CMakeLists like this:
```cmake
add_test(NAME parallel-mpi COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${MPIEXEC_MAX_NUMPROCS}
                                   ${MPIEXEC_PREFLAGS} $<TARGET_FILE:parallel_mpi_test> ${MPIEXEC_POSTFLAGS})

# Add "mpi" label to the test and set a default process number to launch.
set_tests_properties(parallel-mpi PROPERTIES LABELS "mpi" PROCESSORS ${MPIEXEC_MAX_NUMPROCS})
```

The parallel run is launch with:
```bash
# run all test labeled by `mpi
ctest --preset linux-gcc-release -L mpi   --output-junit output-mpi.xml
```
All the other tests can be launched with one line:
```bash
ctest --preset linux-gcc-release -LE mpi   --output-junit output-serial.xml
```

It is interesting because it is independent of the mpi implementation (mpich, openmpi etc...) and use the number of processors detected on the host. 

Note: it's tempting to use the `--parallel` flag to execute serial tests concurrently but Khiops is not thread-safe

### LINKED ISSUE #203: Switch to OpenMPI for Linux packages

### LINKED ISSUE #207: Driver loading causes unwanted warning in jupyter
#### Description
When khiops has a cloud storage driver installed, it dumps a message to stdout.
This message is considered as abnormal by khiops-python - normally all messages are set to the logs - and result in an ugly error when run in a jupyter notebook.

#### Questions/Ideas
- why not reroute this message to traces, debug ?
- how about suppressing it entirely, it would be a 5s fix
- fixing it in khiops-python is feasible but would imply making an exception just for this case

### Context
- Khiops 10.2.0
- khiops-python 10.2.0.0

image

### LINKED ISSUE #216: Change exit status of Khiops
The exit status of MODL and MODL_coclustering  is as follows:
- return 0 when there is no failure
- return 1 when there is a fatal error
- return 2  when there is a user warning/error

This is a problem with open MPI because it adds the following lines when the return code is not 0:

> 
> --------------------------------------------------------------------------
> Primary job  terminated normally, but 1 process returned
> a non-zero exit code. Per user-direction, the job has been aborted.
> --------------------------------------------------------------------------
> --------------------------------------------------------------------------
> mpirun detected that one or more processes exited with non-zero status, thus causing
> the job to be terminated. The first process to do so was:
> 
>   Process name: [[34210,1],0]
>   Exit code:    2
> --------------------------------------------------------------------------

We have to remove the exit code 2, to keep these lines only in case of fatal error.

bruno-at-orange:
Switch to OpenMPI for rpm and deb packages. The MPI implementation is now a parameter of the actions `pack-debian` and `pack-rpm.` The packages and binaries names contain the name of the mpi implementation (openmpi or mpich).

The `run-unit-test` action works indifferently with openMPI or mpich by using ctest (close #23).
However, the `run-standard-test` action works only with binaries built with openmpi.

close #203 
may close the issue #164 as we use openMPI as default mpi implementation

Extra features (unrelated with openmpi):
- Add environment variable to launch a file server process on single host 
- close #216 
- close #207
- cherry pick LearningTestTool from dev


.github/actions/build-khiops/action.yml
```diff
@@ -13,8 +13,22 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Setup MPI
+    - name: Setup MPI (windows)
+      if: ${{ runner.os == 'Windows' }}
```
folmos-at-orange: `${{ }}` is not necessary in `if` statements. You may write
```yaml
if: runner.os == 'Windows'
```

.github/actions/build-khiops/action.yml
```diff
@@ -13,8 +13,22 @@ inputs:
...
       uses: mpi4py/setup-mpi@v1
+    - name: Setup MPI (macOS)
+      if: ${{ runner.os == 'macOS' }}
```
folmos-at-orange: `${{ }}` is not necessary in `if` statements.

.github/actions/build-khiops/action.yml
```diff
@@ -13,8 +13,22 @@ inputs:
...
+      shell: bash
+      run: |
+        brew install mpich
+    - name: Setup MPI (Linux)
+      if: ${{ runner.os == 'Linux' }}
```
folmos-at-orange: `${{ }}` is not necessary in `if` statements.

.github/workflows/pack-debian.yml
```diff
@@ -82,29 +99,38 @@ jobs:
...
           path: artifacts
       - name: Install Khiops core
         run: |
+          export DEBIAN_FRONTEND="noninteractive"
           apt-get update
           dpkg -i ./artifacts/khiops-core* || true
           apt-get -f install -y
       - name: Check Khiops core installation
-        continue-on-error: ${{ matrix.os == 'debian:11' }}
+        #continue-on-error: ${{ matrix.os == 'debian:11' }}
```
folmos-at-orange: If this isn't a problem anymore then it should be removed.

.github/workflows/pack-debian.yml
```diff
@@ -82,29 +99,38 @@ jobs:
...
           khiops -v
           khiops_coclustering -v
       - name: Install Khiops Desktop (with java)
         run: |
-          dpkg -i ./khiops_* || true
+          dpkg -i ./artifacts//khiops_* || true
           apt-get -f install -y
       - name: Check Khiops installation
-        continue-on-error: ${{ matrix.os == 'debian:11' }}
+        #continue-on-error: ${{ matrix.os == 'debian:11' }}
```
folmos-at-orange: If this isn't a problem anymore then it should be removed.

.github/workflows/pack-rpm.yml
```diff
@@ -32,17 +38,22 @@ jobs:
...
+          else
+            echo "MPI_IMPLEMENTATION=${{ github.event.inputs.mpi_implementation }}" >> "$GITHUB_ENV"
+          fi
       - name: Configure CMake
         run: |
           source /etc/profile.d/modules.sh
-          module load mpi/mpich-x86_64
-          echo "MPI_SUFFIX=$MPI_SUFFIX" >> "$GITHUB_ENV"
+          module unload mpi
+          module load mpi/${{ env.MPI_IMPLEMENTATION}}-x86_64
```
folmos-at-orange: If it is in the environment I'd use `$MPI_IMPLEMENTATION`

.github/workflows/pack-rpm.yml
```diff
@@ -32,17 +38,22 @@ jobs:
           echo "ID=$ID" >> "$GITHUB_ENV"
           VERSION_CODENAME=$(echo $PLATFORM_ID | cut -d":" -f2)
           echo "VERSION_CODENAME=$VERSION_CODENAME" >> "$GITHUB_ENV"
+          if [ -z "${{ github.event.inputs.mpi_implementation }}" ] ; 
+          then
+            echo "MPI_IMPLEMENTATION=openmpi" >> "$GITHUB_ENV"
+          else
+            echo "MPI_IMPLEMENTATION=${{ github.event.inputs.mpi_implementation }}" >> "$GITHUB_ENV"
```
folmos-at-orange: A simpler way to do this is
```bash
MPI_IMPLEMENTATION="${{ inputs.mpi_implementation || 'openmpi' }}"
echo "MPI_IMPLEMENTATION=$MPI_IMPLEMENTATION" >> "$GITHUB_ENV"
```
or even

```bash
echo "MPI_IMPLEMENTATION=${{ inputs.mpi_implementation || 'openmpi' }}" >> "$GITHUB_ENV"
```

.github/workflows/pack-rpm.yml
```diff
@@ -88,6 +99,12 @@ jobs:
           echo "ID=$ID" >> "$GITHUB_ENV"
           VERSION=$(echo $PLATFORM_ID | cut -d":" -f2)
           echo "VERSION=$VERSION" >> "$GITHUB_ENV"
+          if [ -z "${{ github.event.inputs.mpi_implementation }}" ] ; 
+          then
+            echo "MPI_IMPLEMENTATION=openmpi" >> "$GITHUB_ENV"
+          else
+            echo "MPI_IMPLEMENTATION=${{ github.event.inputs.mpi_implementation }}" >> "$GITHUB_ENV"
+          fi
```
folmos-at-orange: I'd use any of the simplifications mentioned above.

.github/workflows/run-standard-tests.yml
```diff
@@ -48,24 +48,32 @@ jobs:
         uses: ./.github/actions/build-khiops
         with:
           preset-name: ${{ env.PRESET_NAME }}
-      - name: Cache binaries
-        id: cache-binaries-unix
+      - name: Cache binaries (windows)
+        id: cache-binaries-windows
+        if: ${{ runner.os == 'Windows' }}
```
folmos-at-orange: `if` statements don't require `${{ }}`

.github/workflows/pack-debian.yml
```diff
@@ -33,8 +41,17 @@ jobs:
...
+          then
+            echo "MPI_IMPLEMENTATION=openmpi" >> "$GITHUB_ENV"
+          else
+            echo "MPI_IMPLEMENTATION=${{ github.event.inputs.mpi_implementation }}" >> "$GITHUB_ENV"
+          fi
+
       - name: Configure CMake
         run: |
+          update-alternatives --set mpi /usr/bin/mpicc.${{ env.MPI_IMPLEMENTATION }}
+          update-alternatives --set mpirun /usr/bin/mpirun.${{ env.MPI_IMPLEMENTATION }}
```
folmos-at-orange: If it is in env you may use just `$MPI_IMPLEMENTATION`

.github/workflows/pack-debian.yml
```diff
@@ -82,29 +99,38 @@ jobs:
           source /etc/os-release
           echo "ID=$ID" >> "$GITHUB_ENV"
           echo "VERSION_CODENAME=$VERSION_CODENAME" >> "$GITHUB_ENV"
+          if [ -z "${{ github.event.inputs.mpi_implementation }}" ] ; 
+          then
+            echo "MPI_IMPLEMENTATION=openmpi" >> "$GITHUB_ENV"
+          else
+            echo "MPI_IMPLEMENTATION=${{ github.event.inputs.mpi_implementation }}" >> "$GITHUB_ENV"
+          fi
+
```
folmos-at-orange: You may use one of the simplifications mentioned below.

.github/workflows/run-standard-tests.yml
```diff
@@ -83,35 +91,63 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
-      - name: Restore cached binaries
-        id: restore-binaries
+      - name: Restore cached binaries (windows)
+        if: ${{ runner.os == 'Windows' }}
```
folmos-at-orange: `if` statements dont require `${{ ... }}`

.github/workflows/run-standard-tests.yml
```diff
@@ -48,24 +48,32 @@ jobs:
...
             ${{ github.workspace }}/build/${{ env.PRESET_NAME }}/bin/KNITransfer.exe
             ${{ github.workspace }}/build/${{ env.PRESET_NAME }}/bin/KhiopsNativeInterface.dll
-            ${{ github.workspace }}/build/${{ env.PRESET_NAME }}/lib/libKhiopsNativeInterface.*
             ${{ github.workspace }}/build/${{ env.PRESET_NAME }}/lib/KhiopsNativeInterface.lib
           key: ${{ github.run_id }}-${{ matrix.build-setup.os }}-${{ env.PRESET_NAME
             }}
           fail-on-cache-miss: false
+      - name: Cache binaries (macOS, Linux)
+        id: cache-binaries-unix
+        if: ${{ runner.os != 'Windows' }}
```
folmos-at-orange: `if` statements don't require `${{ }}`

.github/workflows/run-standard-tests.yml
```diff
@@ -83,35 +91,63 @@ jobs:
...
         run: |
-          echo "TEST_PY=test/LearningTest/cmd/python/test_khiops.py" >> "$GITHUB_ENV"
-          echo "APPLY_PY=test/LearningTest/cmd/python/apply_command.py" >> "$GITHUB_ENV"
+          echo "TEST_PY=test/LearningTestTool/py/kht_test.py" >> "$GITHUB_ENV"
+          echo "APPLY_PY=test/LearningTestTool/py/kht_apply.py" >> "$GITHUB_ENV"
+          echo "COLLECT_PY=test/LearningTestTool/py/kht_collect_results.py" >> "$GITHUB_ENV"
           echo "BIN_PATH=${{ github.workspace }}/build/${{ env.PRESET_NAME }}/bin" >> "$GITHUB_ENV"
-      - name: Setup MPI
+      - name: Setup MPI (windows)
+        if: ${{ runner.os == 'Windows' }}
```
folmos-at-orange: `if` statements don't require `${{ }}`

.github/workflows/run-standard-tests.yml
```diff
@@ -83,35 +91,63 @@ jobs:
...
         uses: mpi4py/setup-mpi@v1
+      - name: Setup MPI (macOS)
+        if: ${{ runner.os == 'macOS' }}
```
folmos-at-orange: `if` statements don't require `${{ }}`

.github/workflows/run-standard-tests.yml
```diff
@@ -83,35 +91,63 @@ jobs:
...
+        shell: bash
+        run: |
+          brew install mpich
+      - name: Setup MPI (Linux)
+        if: ${{ runner.os == 'Linux' }}
```
folmos-at-orange: `if` statements don't require `${{ }}`

.github/workflows/run-standard-tests.yml
```diff
@@ -126,59 +162,61 @@ jobs:
...
+          MSG="${MSG//'%'/'%25'}"
+          MSG="${MSG//$'\n'/'%0A'}"
+          MSG="${MSG//$'\r'/'%0D'}"
+          if grep -q -e "error" -e "The test has not been launched" results/errors.txt ; then
+            echo "::error::$MSG"
             false
           fi
-          if grep -qr "The test has not been launched" --include="errors.txt" test/LearningTest/ ; then
-            echo "::error::Test not launched"
+          if [ "${{ github.event.inputs.warning_as_error }}" == "true" ] && grep -q "warning" results/errors.txt  ; then
```
folmos-at-orange: You may use `inputs.warnings_as_error` instead of `github.event.inputs.warnings_as_error`

test/LearningTestTool/py/_kht_utils.py
```diff
@@ -152,6 +152,32 @@ def parent_dir_name(dir_path, depth):
...
+
+def is_valid_tool_full_exe_name(tool_exe_name):
+    """Indique si le nom du binaire fait partie des noms valides, avec prise
+    en compte des suffixes mpi pour les exe paralellisable"""
+
+    if tool_exe_name in kht.TOOL_EXE_NAMES.values():
+        return True
+    if extract_tool_exe_name(tool_exe_name) in kht.PARALLEL_TOOL_NAMES:
+        return True
+    return False
```
folmos-at-orange: Maybe simpler with
```python
return (tool_exe_name in kht.TOOL_EXE_NAMES.values() or extract_tool_exe_name(tool_exe_name) in kht.PARALLEL_TOOL_NAMES)
```

folmos-at-orange (changes requested):
Mostly cosmetic stuff.

One question: It is possible to have just one binary which can be linked to both OpenMPI and MPICH ?

_Edited_


And another two questions: 
- Did we test for performance degradations ?
- It is pertinent to migrate to openmpi for conda packages ?

.github/actions/build-khiops/action.yml
bruno-at-orange: done

.github/workflows/pack-rpm.yml
bruno-at-orange: OK, nice...

folmos-at-orange:
I added two commits.

Good news:
The Linux packages pass the LearningTest battery.

Bad news:
The conda packages are very difficult make work with openmpi so I left them with MPICH. More details in the internal email on the subject.

src/Learning/KWUtils/KWKhiopsVersion.h
```diff
@@ -10,7 +10,7 @@
 // dans le TaskManager de Windows (par exemple)
 
 // Version de Khiops
-#define KHIOPS_VERSION KHIOPS_STR(10.2.1)
+#define KHIOPS_VERSION KHIOPS_STR(10.2.2-b0)
```
folmos-at-orange: To be SemVer compatible the version should be `10.2.2-b.0` (a period `.` between the `b` and the `0`)

folmos-at-orange (approved):
I think the version should be changed. See comment.

If it is possible, could we have a "no suffix" generation even if MPI is available ? In conda we do not need the suffixes, and it simplifies the lookup for the executable. At the moment I just rename it after the build.

PR #238, opened 2024-04-15
force gcc optimization to O1 in KWDGPODiscretizer

### LINKED ISSUE #163: Segmentation fault of coclustering on debian 12
#### Description
Khiops Coclustering fails with a segmentation fault on `LearningTest/TestCoclustering/Standard/Iris/ `
:warning:  It is a **non-reproductible bug**

### Context
- Khiops version 10.2.0
- Debian 12

bruno-at-orange:
With gcc 12.2.0 (on debian 12), gcc seems too aggressive. The only way to fix a segmentation fault is to force the optimization level to O1 instead of O2.

The segmentation fault occurred while accessing the variable cell1 which is NULL. By adding the following (unnecessary) line, the segmentation fault vanishes: 
```cpp
if (cell1==NULL or cell2==NULL) exit(1);
```
With O1 optimization level, this line is not necessary to cope with this segmentation fault
close #163

### DIFF
--- a/.github/workflows/pack-debian.yml
+++ b/.github/workflows/pack-debian.yml
@@ -108,7 +108,7 @@ jobs:
           khiops -v
           khiops_coclustering -v
       - name: Test Khiops installation
-        continue-on-error: ${{ matrix.os == 'debian:11' || matrix.os == 'debian:12' }}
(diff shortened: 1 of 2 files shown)


marcboulle (approved):
LGTM
Cette correction sera à valider définitivement en relançant tous les test de LearningTest

PR #239, opened 2024-04-16
force gcc optimization to O1 in KWDGPODiscretizer

bruno-at-orange:
With gcc 12.2.0 (on debian 12), gcc seems too agressive. The only way to fix a segmentation fault is to force the optimization level to O1 instead of O2.

The segmentation fault occured while accessing the variable cell1 which is NULL. By adding the following (unnecessary) line, the segmentation fault vanishes: if (cell1==NULL or cell2==NULL) exit(1);

cherry pick from dev commit bfc01f05727fc1b649105edd351a52d70805ba4b PR #163

### DIFF
--- a/.github/workflows/pack-debian.yml
+++ b/.github/workflows/pack-debian.yml
@@ -108,7 +108,7 @@ jobs:
           khiops -v
           khiops_coclustering -v
       - name: Test Khiops installation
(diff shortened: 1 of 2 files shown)


marcboulle (approved):
LGTM

PR #245, opened 2024-04-17
remove post-installations script from khiops packages

### LINKED ISSUE #221: Rockylinux packages issue with java version and posint script
There are two issues with rpm packages:
- the post-installation script is run before the dependencies installation (at least on local install). Consequently, java is not found and JAVA_HOME is not set in the `khiops-env ` script and so the GUI is not launched.
- the jdk used to build the jars is **java-17-openjdk-devel** whereas the default java is 1.8. We have the message:
> Exception in thread "main" java.lang.UnsupportedClassVersionError: normGUI/engine/GUIObject has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0

bruno-at-orange:
Java configuration is in khiops-env in the package khiops-core. It is not updated by khiops package anymore. The trick is that this configuration is triggered by the presence of khiops.jar

On the docker files, We pin the jdk version to the smallest one on each distro (it matches most of the time the default-jdk package) The idea is to build norm.jar with the oldest jdk to have a jar compatible with all jre. Ideally we should build with jdk 1.8 on all distros but this package is not available everywhere.


bruno-at-orange:
I change the docker image name to khiopsdev-**test**-os in order to launch CI/CD without interfering with the other branches. Of course I will switch back to the original name and remove the test images before the merge.

packaging/linux/common/khiops-env
```diff
@@ -65,13 +65,13 @@ then
...
-# JAVA_HOME automatic setting
-
-# Set java environment for GUI if Khiops GUI is installed
-if [[ -f "/usr/share/khiops/khiops.jar" && -n "$JAVA_HOME" ]]; then
-   KHIOPS_CLASSPATH=/usr/share/khiops/norm.jar:/usr/share/khiops/khiops.jar
-   KHIOPS_JAVA_PATH=$(find -L "$JAVA_HOME" -name libjvm.so | xargs dirname)
+# Set java environment if Khiops GUI is installed
+if [ -f "/usr/share/khiops/khiops.jar" ]; then
+    if [ -z "$JAVA_HOME" ]; then 
+        JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:bin/java::" )
```
popescu-v, line 71: As this computation would be done dynamically at each `khiops` launch (as `khiops` sources `khiops-env`), then an issue might occur IMHO, if Java is not installed for some reason, or otherwise removed by the user at some time after installing Khiops.

Indeed, in this case `/usr/bin/java` would not exist, hence this line would set `JAVA_HOME` to `/usr/`. Hence, this incurs the risk of failing later on in this script, when trying to set `KHIOPS_JAVA_PATH`, as this would launch `find` inside `/usr/` and either fail with "Permission denied" if the user is not root, or with "dirname: missing operand" otherwise (as `libjvm.so` would not be found).

Is this intended?

popescu-v (commented):
There is one issue to clarify / settle IMHO (see comment).

packaging/linux/common/khiops-env
bruno-at-orange: what do you think about this:
```bash
# Set java environment if Khiops GUI is installed
if [ -f "/usr/share/khiops/khiops.jar" ]; then
    if [ -z "$JAVA_HOME" ] && [ -f /usr/bin/java ]; then
        JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:bin/java::" )
    fi
    if [ -n "$JAVA_HOME" ]; then
        KHIOPS_JAVA_PATH=$(find -L "$JAVA_HOME" -name libjvm.so | xargs dirname)
        KHIOPS_CLASSPATH=/usr/share/khiops/norm.jar:/usr/share/khiops/khiops.jar
    else
        echo "Java not found. Khiops cannot launch a graphical interface. Please install a Java Runtime Environment (>= 1.8)" 
    fi
fi
```
I think it handles the problem

popescu-v: IMHO, there is still one situation that I think we need to account for and also send a (specific?) warning message:

if the user sets `JAVA_HOME` to some bogus value before launching Khiops, e.g. the user sets `JAVA_HOME` to `/foo/bar/baz` which does not exist, then the second `if` would execute directly, bringing us back to the previous issue (bogus `JAVA_HOME`).

Perhaps at least checking that the directory pointed to by `JAVA_HOME` exists? And if not, warn the user specifically that `JAVA_HOME` is bogus? Or perhaps use the same warning message that Java cannot be found (but at least do not fail later on?)

bruno-at-orange: Yes indeed, with this new test I think it would be OK
```bash
# Set java environment if Khiops GUI is installed
if [ -f "/usr/share/khiops/khiops.jar" ]; then
    if [ -z "$JAVA_HOME" ] && [ -f /usr/bin/java ]; then 
        JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:bin/java::" )
    fi
    if [ ! -d "$JAVA_HOME" ]; then
        echo "the JAVA_HOME directory doesn't exist (${JAVA_HOME})"
        JAVA_HOME=""
    fi
    if [ -n "$JAVA_HOME" ]; then
        KHIOPS_JAVA_PATH=$(find -L "$JAVA_HOME" -name libjvm.so | xargs dirname)
        KHIOPS_CLASSPATH=/usr/share/khiops/norm.jar:/usr/share/khiops/khiops.jar
    else
        echo "Java not found. Khiops cannot launch a graphical interface. Please install a Java Runtime Environment (>= 1.8)" 
    fi
fi
```

popescu-v: Yes, this seems to cover all cases IMHO.

I would just start the non-existing `JAVA_HOME` directory warning message with a capital letter, and end it with a full stop. I would also end the second warning message with a full stop.

popescu-v: Also, I think `unset JAVA_HOME` also works, as, in the context of the script, it would have the same effect as assigning it to the empty string.

bruno-at-orange: All your comments are in the new commit !

popescu-v (approved):
LGTM.

PR #252, opened 2024-05-07
Switch openmpi

bruno-at-orange:
cherry pick from dev-v10:
- Add environment variable to launch a file server process on single host
- switch to openMPI
- remove post-installations script from khiops packages

### DIFF
--- a/.github/actions/build-khiops/action.yml
+++ b/.github/actions/build-khiops/action.yml
@@ -13,12 +13,26 @@ inputs:
 runs:
   using: composite
   steps:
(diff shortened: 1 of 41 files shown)


bruno-at-orange:
I pinned the version of the macos runner to **macos-13** because the **macos-latest**  became **macos-14** during the week-end. And there is an instability in LearningTest/TestKhiops/BugMPIWithErrors.
Noted in issue #255

marcboulle (approved):
LGTM

PR #256, opened 2024-05-13
Add new KNITransfer package

### LINKED ISSUE #188: Put  KNITransfer in the releases for testing
#### Description
Currently, the internal testing binary for KNI, `KNITransfer` isn't put in the releases action. We should put it so it is available for the `LearningTest` infra.

#### Questions/Ideas
- Is a real package necessary? or a simple zip suffices.
  - 04/04/24: Not necessary but easier to make.

bruno-at-orange:
Add package for the binary **KNITransfer**:
- deb and rpm for linux
- zip for windows

These packages are uploaded  to the _github releases_  on tags


marcboulle (approved):
LGTM
Juste supprimer le WIP du titre du dernier commit

PR #269, opened 2024-05-23
Improve packaging

### LINKED ISSUE #257: CMake warning on Windows
When configuring project on windows, cmake fails to build a symlink for the MODL target, producing a warning.
This symlink is needed only for Linux packages. We should condition the install with os names.

bruno-at-orange:
Improve packaging on Linux
fix #257


marcboulle (approved):
LGTM

PR #272, opened 2024-05-24
Improve Linux packaging

bruno-at-orange:
* Install KHIOPS and KHIOPS_CORE components only for linux (suppress cmake warning on windows)

* Improve generation of khiops-env with cmake (remove absolute path of mpiexec)
* The khiops-core-openmpi package provides the virtual package khiops-core. It improves dependency managment.
* Suppress output of OpenMPI in stdout


marcboulle (approved):
LGTM

PR #286, opened 2024-06-04
fix  wrong driver list in the status info

bruno-at-orange:
* Display all drivers in Show ShowSystemInformation

In the ShowSystemInformation method (it corresponds to the -s flag) we diplay the list of all drivers.

We print the list of the environment variables if they are defined.

* Add fatal error when KHIOPS_MEMORY_LIMIT is too small

if KHIOPS_MEMORY_LIMIT is inferior than the one required by khiops. It was simply ignored. I add a Fatal Error to inform the user

Minor corrections/update in the khiops-env scripts

cherry-pick from dev commit 9af20ba103f9c10ad59c10b5ca3f7280e249f097


marcboulle (approved):
LGTM

PR #290, opened 2024-06-06
Fix windows conda build

folmos-at-orange:
_Cherry-pick of the same fix for `dev-v10`, issue #271_

The windows-2022 runner does not contain the VS2017 tools and runtime to build the conda khiops package. We will use windows-2019 until the conda-forge packages use a more recent VS.


bruno-at-orange (approved):
LGTM

PR #306, opened 2024-06-17
Add conda package for kni and KNITransfer

bruno-at-orange:
In `LearningTestRunner`, the new python api (kht_test) runs systematically the tests for Khiops, Khiops_coclustering and **KNI**. It fails if KNITransfer is missing. Therefore I added 2 new conda packages (as sub-packages): kni and kni-transfer.

A second commit concerns `kht_test.py`. The origin of the problem is that linux packages use **openmpi** and conda package use **mpich**. `kht_test` is now compliant with both: the flags specific to openmpi are translate in environment variables.

### DIFF
--- a/packaging/conda/bld.bat
+++ b/packaging/conda/bld.bat
@@ -3,11 +3,13 @@ REM Echo all output
 
 REM Build the Khiops binaries
 cmake --preset windows-msvc-release -DBUILD_JARS=OFF -DTESTING=OFF
-cmake --build --preset windows-msvc-release --parallel --target MODL MODL_Coclustering
+cmake --build --preset windows-msvc-release --parallel --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer 
 
 REM Copy the MODL binaries to the Conda PREFIX path
 mkdir %PREFIX%\bin
 copy build\windows-msvc-release\bin\MODL.exe %PREFIX%\bin
 copy build\windows-msvc-release\bin\MODL_Coclustering.exe %PREFIX%\bin
+copy build\windows-msvc-release\bin\KNITransfer.exe %PREFIX%\bin
+copy build\windows-msvc-release\bin\KhiopsNativeInterface.dll %PREFIX%\lib
 
 if errorlevel 1 exit 1
--- a/packaging/conda/build.sh
+++ b/packaging/conda/build.sh
@@ -15,10 +15,12 @@ fi
 cmake --fresh --preset $CMAKE_PRESET -DBUILD_JARS=OFF -DTESTING=OFF
 
 # Build MODL and MODL_Coclustering
-cmake --build --preset $CMAKE_PRESET --parallel --target MODL MODL_Coclustering
+cmake --build --preset $CMAKE_PRESET --parallel --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer 
 
-# Move the MODL binaries to the Conda PREFIX path
+# Move the binaries to the Conda PREFIX path
 mv ./build/$CMAKE_PRESET/bin/MODL* "$PREFIX/bin"
+mv ./build/$CMAKE_PRESET/bin/KNITransfer* "$PREFIX/bin"
+mv ./build/$CMAKE_PRESET/lib/libKhiopsNativeInterface* "$PREFIX/lib"
 
 # Custom rpath relocation and signing executables for macOS in arm64
 #
--- a/packaging/conda/meta.yaml
+++ b/packaging/conda/meta.yaml
@@ -1,5 +1,5 @@
 package:
-  name: khiops-core
+  name: khiops-binaries
   version: {{ os.environ.get('KHIOPS_VERSION') }}
 
 source:
@@ -37,15 +37,57 @@ requirements:
     - {{ mpi }}
 
 outputs:
+
+  # khiops-core package
   - name: khiops-core
+    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    requirements:
+      build:
+        - cmake
+        - ninja
+        - {{ compiler('cxx') }}
+      host:
+        - {{ mpi }}
+      run:
+        - {{ mpi }}
+    files:
+      - bin/MODL*
     test:
       commands:
-        - MODL -v
-        - MODL_Coclustering -v
+        - MODL -s
+        - MODL_Coclustering -s
 
+  # kni package (do not need khiops-core as a runtime dependency)
+  - name: kni
+    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    files: 
+      - lib/libKhiopsNativeInterface.so* # [linux]
+      - lib/libKhiopsNativeInterface*.dylib # [osx]
+      - lib/KhiopsNativeInterface.dll # [win]
+    requirements:
+      build:
+        - cmake
+        - ninja
+        - {{ compiler('cxx') }}
+  
+  # kni-transfer package (designed only to test kni)
+  - name: kni-transfer
+    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    requirements:
+      build:
+        - cmake
+        - ninja
+        - {{ compiler('cxx') }}
+      run:
+        - kni 
+    files: 
+      - bin/KNITransfer.exe # [win]
+      - bin/KNITransfer # [linux or osx]
+      
 about:
   home: https://khiops.org
   license: BSD+3-clause
   summary: Khiops is an AutoML suite for supervised and unsupervised learning
   doc_url: https://khiops.org
   dev_url: https://github.com/khiopsml/khiops
+
--- a/test/LearningTestTool/py/kht_test.py
+++ b/test/LearningTestTool/py/kht_test.py
@@ -311,14 +311,27 @@ def evaluate_tool_on_test_dir(
                 utils.remove_file(file_path)
 
         # khiops en mode expert via une variable d'environnement
-        os.putenv(kht.KHIOPS_EXPERT_MODE, "true")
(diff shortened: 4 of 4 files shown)


marcboulle (approved):
Ok pour kht_test.py

Juste une question:
- Linux, c'est maintenant systématiquement du OpenMPI?
- pas de possibilité de switcher entre MPICH et OpenMPI?

En fait, il suffit de compléter le commentaire pour bien préciser que les deux versions sont possible, puisque l'une est paramètres par des variables d'environnement (OpenMPI), que l'autre ignore.

test/LearningTestTool/py/kht_test.py
```diff
@@ -320,6 +320,16 @@ def evaluate_tool_on_test_dir(
...
 
+        # Ajout de variables d'environements propres a OpenMPI, elles remplacent les parametres
+        if platform.system() == "Linux":
+            # Supprime les traces en cas d'erreur fatale de khiops. Option --quiet
+            os.putenv("OMPI_MCA_orte_execute_quiet", "true")
+            # permet de lancer plus de processus qu'il n'y a de coeurs. Option --oversubscribe
+            os.putenv("OMPI_MCA_rmaps_base_oversubscribe", "true")  #
+            # permet de lancer en tant que root. Option --allow-run-as-root
+            os.putenv("OMPI_ALLOW_RUN_AS_ROOT", "1")
+            os.putenv("OMPI_ALLOW_RUN_AS_ROOT_CONFIRM", "1")
```
popescu-v: IMHO, it is preferrable to replace the `os.putenv(VAR, value)` with `os.environ[VAR] = value`, because, as stated in the official documentation https://docs.python.org/3/library/os.html#os.putenv, `os.putenv` updates the environment, but does not update `os.environ` (as `os.environ` is updated at import time or via direct assignment), whereas the direct assignment updates both (as the direct assignment calls `os.putenv` behind the scenes anyway).

packaging/conda/bld.bat
```diff
@@ -3,11 +3,13 @@ REM Echo all output
...
 cmake --preset windows-msvc-release -DBUILD_JARS=OFF -DTESTING=OFF
-cmake --build --preset windows-msvc-release --parallel --target MODL MODL_Coclustering
+cmake --build --preset windows-msvc-release --parallel --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer 
 
 REM Copy the MODL binaries to the Conda PREFIX path
 mkdir %PREFIX%\bin
 copy build\windows-msvc-release\bin\MODL.exe %PREFIX%\bin
 copy build\windows-msvc-release\bin\MODL_Coclustering.exe %PREFIX%\bin
+copy build\windows-msvc-release\bin\KNITransfer.exe %PREFIX%\bin
+copy build\windows-msvc-release\bin\KhiopsNativeInterface.dll %PREFIX%\bin
```
popescu-v: Do we need to add the DLL to `bin`? I would have added it to `lib` instead, as it is a library and it would seem clearer to me to have it there.

packaging/conda/build.sh
```diff
@@ -15,10 +15,12 @@ fi
...
 
 # Build MODL and MODL_Coclustering
-cmake --build --preset $CMAKE_PRESET --parallel --target MODL MODL_Coclustering
+cmake --build --preset $CMAKE_PRESET --parallel --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer 
 
-# Move the MODL binaries to the Conda PREFIX path
+# Move the binaries to the Conda PREFIX path
 mv ./build/$CMAKE_PRESET/bin/MODL* "$PREFIX/bin"
+mv ./build/$CMAKE_PRESET/bin/KNITransfer* "$PREFIX/bin"
+mv ./build/$CMAKE_PRESET/lib/libKhiopsNativeInterface* "$PREFIX/bin"
```
popescu-v: Do we need to add the `*.so`s to `bin`? I would have added them to `lib` instead, as they are libraries and it would seem clearer to me to have them there.

packaging/conda/meta.yaml
```diff
@@ -1,5 +1,5 @@
 package:
-  name: khiops-core
+  name: khiops
```
popescu-v: The Python Conda package is also named `khiops` (and published as such) and it depends on the Conda `khiops-core` package. Nonetheless, I can see that in the `outputs` section we separate the packages and find the expected `khiops-core` package. But to avoid any possible (developer?) confusion, I would tend to give the recipe another name,  like `khiops-binaries` or alike.

packaging/conda/meta.yaml
```diff
@@ -36,16 +36,57 @@ requirements:
   run:
     - {{ mpi }}
 
+# requirements:
+#   run:
+#     - khiops-core
+#     - kni
+#     - kni-transfer
+
```
popescu-v: This is not needed, right?

packaging/conda/meta.yaml
```diff
@@ -36,16 +36,57 @@ requirements:
...
     test:
       commands:
         - MODL -v
         - MODL_Coclustering -v
+  - name: kni
+    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    files: 
+      - bin/libKhiopsNativeInterface* # [linux or osx]
+      - bin/KhiopsNativeInterface.dll # [win]
+    requirements:
```
popescu-v, line 67: I would add a comment here that KNI does not need  `khiops-core` as a runtime dependency.

packaging/conda/meta.yaml
```diff
@@ -36,16 +36,57 @@ requirements:
...
+    files:
+      - bin/MODL*
     test:
       commands:
         - MODL -v
         - MODL_Coclustering -v
+  - name: kni
+    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    files: 
+      - bin/libKhiopsNativeInterface* # [linux or osx]
```
popescu-v: Why the glob here? Is the extension on MacOS different from Linux?

packaging/conda/meta.yaml
```diff
@@ -36,16 +36,57 @@ requirements:
...
+    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    requirements:
+      build:
+        - cmake
+        - ninja
+        - {{ compiler('cxx') }}
+      run:
+        - kni 
+    files: 
+      - bin/KNITransfer*
```
popescu-v: Why the glob here? Is there for the Windows `.exe` extension? If so, then why not put both, with `[linux or osx]` vs. `[win]`?

popescu-v (changes requested):
There are some pending comments.

test/LearningTestTool/py/kht_test.py
bruno-at-orange: done

packaging/conda/meta.yaml
bruno-at-orange: ok

packaging/conda/meta.yaml
bruno-at-orange: yep

packaging/conda/meta.yaml
bruno-at-orange: done

packaging/conda/meta.yaml
bruno-at-orange: yes an MacOS, I think it is dylib
And on Linux, with cmake with have automatically one file and 2 symlinks:
```
libKhiopsNativeInterface.so -> libKhiopsNativeInterface.so.10.2.2-b.2*
libKhiopsNativeInterface.so.10 -> libKhiopsNativeInterface.so.10.2.2-b.2*
 ```

File: packaging/conda/meta.yaml (hunk 8 above)
Comment (bruno-at-orange) [reply packaging/conda/meta.yaml]: ok

File: packaging/conda/meta.yaml (hunk 7 above)
Comment (bruno-at-orange) [reply packaging/conda/meta.yaml]: Update: on macOS the versioning follows libname1.2.3.dylib [doc here](https://docstore.mik.ua/orelly/unix3/mac/ch05_04.htm)

File: packaging/conda/bld.bat (hunk 2 above)
Comment (bruno-at-orange) [reply packaging/conda/bld.bat]: I agree, I was lazy. They are in bin now...

File: packaging/conda/build.sh (hunk 3 above)
Comment (bruno-at-orange) [reply packaging/conda/build.sh]: ok

File: test/LearningTestTool/py/kht_test.py [hunk 9]
```diff
@@ -311,14 +311,37 @@ def evaluate_tool_on_test_dir(
...
+
+        # Ajout de variables d'environements propres a OpenMPI, elles remplacent les parametres
+        if platform.system() == "Linux":
+            # Supprime les traces en cas d'erreur fatale de khiops. Option --quiet
+            os.putenv("OMPI_MCA_orte_execute_quiet", "true")
+            # permet de lancer plus de processus qu'il n'y a de coeurs. Option --oversubscribe
+            os.putenv("OMPI_MCA_rmaps_base_oversubscribe", "true")  #
+            # permet de lancer en tant que root. Option --allow-run-as-root
+            os.putenv("OMPI_ALLOW_RUN_AS_ROOT", "1")
+            os.putenv("OMPI_ALLOW_RUN_AS_ROOT_CONFIRM", "1")
```
popescu-v: s/os.putenv(VAR, value)/os.environ[VAR] = value/

packaging/conda/meta.yaml
popescu-v: OK, and OK with your solution with the `*.dylib` glob.

popescu-v:
LGTM.

popescu-v (approved):

PR #311, opened 2024-06-21
Add Java Temurin to windows installer

bruno-at-orange:
Use Eclipse Temurin JRE in the windows installer. The installer is no longer installed on the system while the Khiops installation. It is simply copyed on KHIOPS_HOME. We're no longer dependent on (de)installation and updates of java on the system. And there is no need to look into the windows register.

cherry-pick from dev-v10 8aa8cc3a2ed87f3e6a82696d4142098a3fc42e0f


marcboulle (approved):
LGTM

Issue #307, opened 2024-06-18
OpenMPI-related error on Rocky Linux


popescu-v:
#### Description
When verifying the Khiops installation on a Rocky 8 or Rocky 9 Docker container via `khiops -s` or `khiops -v` , we obtain:

```sh
Open MPI failed an OFI Libfabric library call (fi_endpoint).  This is highly
unusual; your job may behave unpredictably (and/or abort) after this.

  Local host: ...
  Location: mtl_ofi_component.c:513
  Error: Invalid argument (22)
--------------------------------------------------------------------------
Khiops 10.2.2-b.3
```
See:
- https://github.com/KhiopsML/khiops/actions/runs/9468751070/job/26085981246 ; or
- https://github.com/KhiopsML/khiops-python/actions/runs/9566989588/job/26374196376#step:3:10

However, the CIs are green because `khiops -s` and `khiops -v` respectively return 0 nonetheless and only display this error as a warning.

#### Questions/Ideas
- This seems related to: https://github.com/ofiwg/libfabric/issues/6710
  - A potential (at least temporary) fix might be to set `PSM3_MULTI_EP=1` (see https://github.com/ofiwg/libfabric/issues/6710#issuecomment-831119595)
- We need to check whether:
   - the aforementioned fix makes the error go away
   - this error makes OpenMPI unusable or yields performance loss
 - Apparently the libfabric issue has been fixed in https://github.com/ofiwg/libfabric/pull/6737 (merged as part of https://github.com/ofiwg/libfabric/pull/6877), but chances are that the libfabric versions that are available on Rocky 8 and 9 are prior to this merge.

### Context
- Khiops version: 10.2.2-b.3
- OS description: Rocky 8 and Rocky 9 on the GitHub runners

popescu-v:
Adding the aforementioned fix silences the reported error, but triggers another error:
```sh
ORTE_ERROR_LOG: Out of resource in file util/show_help.c at line 507
```
See https://github.com/KhiopsML/khiops-python/actions/runs/9568000054/job/26377335252#step:3:18

See also (perhaps relevant) issue: https://github.com/open-mpi/ompi/issues/5798

PR #313, opened 2024-06-24
307 openmpi related error on rocky linux

bruno-at-orange:
* Add new environment variables to khiops-env
To fix the bug with PSM3, we add "PSM3_DEVICES=self"
For explanation, see https://github.com/open-mpi/ompi/issues/11295#issuecomment-1384244610
We also add OMPI_MCA_btl_vader_single_copy_mechanism=none
to remove the mpi message "Read -1, expected 65536, errno = 1" that appears on docker

* Rename github actions test-khiops-install > test-khiops-on-iris
* Add new github action to test khiops installation
The test-khiops-install checks if the output of Khiops in the console is as expected (no mpi or java errors)

### DIFF
--- a/.github/actions/test-khiops-install/action.yml
+++ b/.github/actions/test-khiops-install/action.yml
@@ -1,0 +1,57 @@
+---
+name: Test Khiops Installation
+description: Check the outputs of Khiops in the console
+runs:
+  using: composite
+  # Test if the output of Khiops in the console is as expected (no mpi or java errors)
+  # We expect only one line in stdout+stderr when running 'khiops -v'
+  steps:
+    - name: Print status for debugging
+      shell: bash
+      # We force the process number to 4 because the runner has only 2 proc.
+      # With 2 proc khiops-env switches to serial and mpi is not used.
+      # The env var OMPI_MCA_rmaps_base_oversubscribe is for openmpi it corresponds to
+      # the flag --oversubscribe (Nodes are allowed to be oversubscribed)
+      env:
+        KHIOPS_PROC_NUMBER: 4
+        OMPI_MCA_rmaps_base_oversubscribe: true
+      run: |
+        khiops-env --env
+    - name: Check Khiops output
+      shell: bash
+      env:
+        KHIOPS_PROC_NUMBER: 4
+        OMPI_MCA_rmaps_base_oversubscribe: true
+      run: |
+        khiops -s
+        khiops -v &> output
+        LINE_NUMBER=$(wc -l < output)
+        if [ $LINE_NUMBER -gt 1 ] ;
+        then 
+          echo "::error::Unexpected output in khiops scripts"
+          false
+        fi
+    - name: Check Khiops coclustering output
+      shell: bash
+      run: |-
+        khiops_coclustering -s
+        khiops_coclustering -v &> output
+        LINE_NUMBER=$(wc -l < output)
+        if [ $LINE_NUMBER -gt 1 ] ;
+        then 
+          echo "::error::Unexpected output in khiops_coclustering scripts"
+          false
+        fi
+    - name: Check process number
+      shell: bash
+      env:
+        KHIOPS_PROC_NUMBER: 4
+        OMPI_MCA_rmaps_base_oversubscribe: true
+      run: |-
+        PROC_NUMBER_LIST=($(khiops -s | grep 'Logical processes on system'))
+        PROC_NUMBER=${PROC_NUMBER_LIST[-1]}
+        if [ "$PROC_NUMBER" != "$KHIOPS_PROC_NUMBER" ] ;
+        then
+          echo "::error::Wrong proc number ($PROC_NUMBER vs $KHIOPS_PROC_NUMBER)"
+          false
+        fi
--- a/.github/actions/test-khiops-on-iris/action.yml
+++ b/.github/actions/test-khiops-on-iris/action.yml
@@ -1,5 +1,5 @@
 ---
-name: Test Khiops Installation
+name: Test Khiops on Iris
 description: Tests a Khiops installation on the Iris dataset
 runs:
   using: composite
--- a/.github/actions/test-khiops-on-iris/action.yml
+++ b/.github/actions/test-khiops-on-iris/action.yml
@@ -10,6 +10,7 @@ runs:
       if: runner.os == 'Linux'
       shell: bash
       run: |
+        # Detect Debian based OS
         if [ -d "/etc/apt" ]
         then
           apt-get install -y python3 > /dev/null
@@ -40,13 +41,17 @@ runs:
         fi
     - name: Check Khiops installation
       shell: bash
+      env:
+        KHIOPS_PROC_NUMBER: 4
+        OMPI_MCA_rmaps_base_oversubscribe: true
       run: |
-        export KHIOPS_PROC_NUMBER=4 
-        export OMPI_MCA_rmaps_base_oversubscribe=true
         "$KHIOPS_SCRIPT" -s
         "$KHIOPS_CC_SCRIPT" -s
     - name: Run Khiops tests
       shell: bash
+      env:
+        KHIOPS_PROC_NUMBER: 4
+        OMPI_MCA_rmaps_base_oversubscribe: true
       run: |
         cd test/LearningTest/TestKhiops/Standard/Iris/
         "$KHIOPS_SCRIPT" -b -i test.prm -e results/err.txt
--- a/.github/workflows/pack-debian.yml
+++ b/.github/workflows/pack-debian.yml
@@ -121,9 +121,9 @@ jobs:
           khiops-env --env
           khiops -s
           khiops_coclustering -s
-      - name: Test Khiops installation
+      - name: Test Khiops on Iris dataset
         continue-on-error: ${{ matrix.os == 'debian:11' }}
-        uses: ./.github/actions/test-khiops-install
+        uses: ./.github/actions/test-khiops-on-iris
   test-kni:
     needs: build
     runs-on: ubuntu-latest
--- a/.github/workflows/pack-debian.yml
+++ b/.github/workflows/pack-debian.yml
@@ -108,19 +108,13 @@ jobs:
           dpkg -i ./artifacts/khiops-core* || true
           apt-get -f install -y
       - name: Check Khiops core installation
-        run: |
-          khiops-env --env
-          khiops -s
-          khiops_coclustering -s
+        uses: ./.github/actions/test-khiops-install
       - name: Install Khiops Desktop (with java)
         run: |
           dpkg -i ./artifacts//khiops_* || true
           apt-get -f install -y
       - name: Check Khiops installation
-        run: |
-          khiops-env --env
-          khiops -s
-          khiops_coclustering -s
+        uses: ./.github/actions/test-khiops-install
       - name: Test Khiops on Iris dataset
         continue-on-error: ${{ matrix.os == 'debian:11' }}
         uses: ./.github/actions/test-khiops-on-iris
--- a/.github/workflows/pack-debian.yml
+++ b/.github/workflows/pack-debian.yml
@@ -116,12 +116,10 @@ jobs:
       - name: Check Khiops installation
         uses: ./.github/actions/test-khiops-install
       - name: Test Khiops on Iris dataset
-        continue-on-error: ${{ matrix.os == 'debian:11' }}
         uses: ./.github/actions/test-khiops-on-iris
   test-kni:
     needs: build
     runs-on: ubuntu-latest
-    continue-on-error: true
     defaults:
       run:
         shell: bash
--- a/packaging/install.cmake
+++ b/packaging/install.cmake
@@ -67,10 +67,20 @@ endif()
 
 if(UNIX)
 
-  # replace MPIEXEC MPIEXEC_NUMPROC_FLAG and MPI_IMPL KHIOPS_MPI_EXTRA_FLAG
+  # replace MPIEXEC MPIEXEC_NUMPROC_FLAG and MPI_IMPL KHIOPS_MPI_EXTRA_FLAG ADDITIONAL_EN_VAR
   if("${MPI_IMPL}" STREQUAL "openmpi")
     set(KHIOPS_MPI_EXTRA_FLAG "--allow-run-as-root --quiet")
+    set(ADDITIONAL_EN_VAR "export OMPI_MCA_btl_vader_single_copy_mechanism=none # issue on docker")
+    if(IS_FEDORA_LIKE)
+      set(ADDITIONAL_EN_VAR "${ADDITIONAL_EN_VAR}\nexport PSM3_DEVICES=self # issue one rocky linux")
+    endif()
   endif()
+
+  # Add header comment to the variable definition (if any variable is defined)
+  if(ADDITIONAL_EN_VAR)
+    set(ADDITIONAL_EN_VAR "# Additional variables for MPI\n${ADDITIONAL_EN_VAR}")
+  endif()
+
   configure_file(${PROJECT_SOURCE_DIR}/packaging/linux/common/khiops-env.in ${TMP_DIR}/khiops-env @ONLY
                  NEWLINE_STYLE UNIX)
   configure_file(${PROJECT_SOURCE_DIR}/packaging/linux/debian/khiops-core/postinst.in ${TMP_DIR}/postinst @ONLY
--- a/.github/workflows/pack-nsis.yml
+++ b/.github/workflows/pack-nsis.yml
@@ -129,7 +129,7 @@ jobs:
       - name: Checkout the khiops sources
         uses: actions/checkout@v4
       - name: Test the installation
-        uses: ./.github/actions/test-khiops-install
+        uses: ./.github/actions/test-khiops-on-iris
   test-kni:
     needs: build
     runs-on: windows-2022
--- a/.github/workflows/pack-rpm.yml
+++ b/.github/workflows/pack-rpm.yml
@@ -105,18 +105,12 @@ jobs:
         run: |
           yum install -y ./artifacts/khiops-core*
       - name: Check Khiops core installation
-        run: |
-          khiops-env --env
-          khiops -s
-          khiops_coclustering -s
+        uses: ./.github/actions/test-khiops-install
       - name: Install Khiops Desktop (with java)
         run: |
           yum install -y ./artifacts/khiops-*
-      - name: Check Khiops  installation
-        run: |
-          khiops-env --env
-          khiops -s
-          khiops_coclustering -s
+      - name: Check Khiops installation
+        uses: ./.github/actions/test-khiops-install
       - name: Test Khiops on Iris dataset
         uses: ./.github/actions/test-khiops-on-iris
   test-kni:
--- a/.github/workflows/pack-rpm.yml
+++ b/.github/workflows/pack-rpm.yml
@@ -117,8 +117,8 @@ jobs:
           khiops-env --env
           khiops -s
           khiops_coclustering -s
-      - name: Test Khiops installation
-        uses: ./.github/actions/test-khiops-install
+      - name: Test Khiops on Iris dataset
+        uses: ./.github/actions/test-khiops-on-iris
   test-kni:
     needs: build
     runs-on: ubuntu-latest
--- a/.github/workflows/pack-rpm.yml
+++ b/.github/workflows/pack-rpm.yml
@@ -75,7 +75,6 @@ jobs:
   test:
     needs: build
     runs-on: ubuntu-latest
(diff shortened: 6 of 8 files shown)


popescu-v:
The Temurin-related commit is not related to the scope of the PR (see issue #307), hence it should belong to a different PR IMHO.

bruno-at-orange:
> The Temurin-related commit is not related to the scope of the PR (see issue #307), hence it should belong to a different PR IMHO.

I deleted this commit. I have no clue why it was here.

popescu-v:
> > The Temurin-related commit is not related to the scope of the PR (see issue #307), hence it should belong to a different PR IMHO.
> 
> I deleted this commit. I have no clue why it was here.

Ok.

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,54 @@
 ---
 name: Test Khiops Installation
-description: Tests a Khiops installation on the Iris dataset
+description: Check the outputs of khiops in the console
```
popescu-v: s/khiops/Khiops/

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,54 @@
...
+  # We expect only one line in stdout+stderr when running 'khiops -v'
   steps:
-    # Python install: we don't use the setup-python action because of the following error:
-    # python3: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by python3)
-    - name: Python setup for Linux
-      if: runner.os == 'Linux'
+    - name: Print status for debugging
       shell: bash
+      # We force the process number to 4 because the runner has only 2 proc.
+      # With 2 proc khiops-env switch to serial and mpi is not used.
```
popescu-v: s/switch/switches/

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,54 @@
...
-        fi
-    - name: Check Khiops installation
-      shell: bash
-      run: |
-        export KHIOPS_PROC_NUMBER=4 
         export OMPI_MCA_rmaps_base_oversubscribe=true
-        "$KHIOPS_SCRIPT" -s
-        "$KHIOPS_CC_SCRIPT" -s
-    - name: Run Khiops tests
+        export KHIOPS_PROC_NUMBER=4
```
popescu-v: IMHO, it's better to set these 2 env vars in an `env:` YAML section of the relevant step:

```
env:
  KHIOPS_PROC_NUMBER: 4
  OMPI_MCA_rmaps_base_oversubscribe: true
```

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,54 @@
...
+        khiops-env --env
+    - name: Check Khiops output
       shell: bash
```
popescu-v, line 21: I think these commands also work for the default shell, which is `sh`, which, on Ubuntu, is `dash`. Why do we need `bash` here?

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,54 @@
...
   using: composite
+  # Test if the output of Khiops in the console is as expected (no mpi or java errors)
+  # We expect only one line in stdout+stderr when running 'khiops -v'
   steps:
-    # Python install: we don't use the setup-python action because of the following error:
-    # python3: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by python3)
-    - name: Python setup for Linux
-      if: runner.os == 'Linux'
+    - name: Print status for debugging
       shell: bash
```
popescu-v, line 10: I don't think we need `bash` here; the default `sh` (`dash` on Ubuntu) shell should suffice IMHO (see the other comment on the same topic).

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,54 @@
...
+        khiops-env --env
+    - name: Check Khiops output
       shell: bash
       run: |
-        cd test/LearningTest/TestKhiops/Standard/Iris/
-        "$KHIOPS_SCRIPT" -b -i test.prm -e results/err.txt
-        cd -
-    - name: Run Khiops Coclustering tests
+        export OMPI_MCA_rmaps_base_oversubscribe=true
+        export KHIOPS_PROC_NUMBER=4
```
popescu-v: Add these to an `env:` section of this step.

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,54 @@
...
-        echo "Check test results"
-        cd test/LearningTestTool/py/
-        $PYTHON kht_test.py ../../LearningTest/TestKhiops/Standard/Iris check
-        $PYTHON kht_test.py ../../LearningTest/TestCoclustering/Standard/Iris check
-        $PYTHON kht_apply.py ../../LearningTest/TestKhiops/Standard/Iris errors | tee /tmp/khiops-log.txt
-        $PYTHON kht_apply.py ../../LearningTest/TestCoclustering/Standard/Iris errors | tee /tmp/coclustering-log.txt
-        if (grep -q error /tmp/khiops-log.txt || grep -q error /tmp/coclustering-log.txt); then
-          echo "::error::Errors in Khiops run"
+        export OMPI_MCA_rmaps_base_oversubscribe=true
+        export KHIOPS_PROC_NUMBER=4
```
popescu-v: Add these 2 env vars to `env:` section (see comments on this topic above).

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,54 @@
...
+        khiops_coclustering -s
+        khiops_coclustering -v &> output
+        LINE_NUMBER=$(wc -l < output)
+        if [ $LINE_NUMBER -gt 1 ] ;
+        then 
+          echo "::error::Unexpected output in khiops_coclustering scripts"
+          false
+        fi
+    - name: Check process number
       shell: bash
```
popescu-v, line 44: I think `sh` suffices (see the other comments on the same topic).

.github/actions/test-khiops-on-iris/action.yml
```diff
@@ -0,0 +1,74 @@
...
+name: Test Khiops on Iris
+description: Tests a Khiops installation on the Iris dataset
+runs:
+  using: composite
+  steps:
+    # Python install: we don't use the setup-python action because of the following error:
+    # python3: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by python3)
+    - name: Python setup for Linux
+      if: runner.os == 'Linux'
+      shell: bash
```
popescu-v, line 11: Why `bash` instead of the default `sh` ?

.github/actions/test-khiops-install/action.yml
bruno-at-orange: done.
I wanted to do this one time for the action as global variable but I think it is not possible for actions (only workflows)

.github/actions/test-khiops-install/action.yml
bruno-at-orange: on windows ?

.github/actions/test-khiops-on-iris/action.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: |
+        if [ -d "/etc/apt" ]
```
popescu-v, line 14: Is  this a robust enough way of sorting out the Ubuntu / Rocky dichotomy? What would prevent some user from creating the `/etc/apt` directory on a Rocky OS?

I would have relied on the relevant env vars `/etc/os-release` instead, for this.

.github/actions/test-khiops-install/action.yml
bruno-at-orange: done

.github/actions/test-khiops-install/action.yml
bruno-at-orange: done

.github/actions/test-khiops-on-iris/action.yml
```diff
@@ -0,0 +1,74 @@
...
+          echo "KHIOPS_CC_SCRIPT=$KHIOPS_CC_SCRIPT" >> "$GITHUB_ENV"
+        else
+          echo "KHIOPS_SCRIPT=khiops" >> "$GITHUB_ENV"
+          echo "KHIOPS_CC_SCRIPT=khiops_coclustering" >> "$GITHUB_ENV"
+        fi
+    - name: Check Khiops installation
+      shell: bash
+      run: |
+        export KHIOPS_PROC_NUMBER=4 
+        export OMPI_MCA_rmaps_base_oversubscribe=true
```
popescu-v: Add these to the `env:` step section.

.github/actions/test-khiops-install/action.yml
bruno-at-orange: done

.github/actions/test-khiops-install/action.yml
bruno-at-orange: done

.github/actions/test-khiops-on-iris/action.yml
bruno-at-orange: done

.github/actions/test-khiops-on-iris/action.yml
```diff
@@ -0,0 +1,74 @@
...
+        "$KHIOPS_SCRIPT" -s
+        "$KHIOPS_CC_SCRIPT" -s
+    - name: Run Khiops tests
+      shell: bash
+      run: |
+        cd test/LearningTest/TestKhiops/Standard/Iris/
+        "$KHIOPS_SCRIPT" -b -i test.prm -e results/err.txt
+        cd -
+    - name: Run Khiops Coclustering tests
+      shell: bash
```
popescu-v, line 55: Why `bash` ?

.github/actions/test-khiops-on-iris/action.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: |
+        cd test/LearningTest/TestCoclustering/Standard/Iris/
+        "$KHIOPS_CC_SCRIPT" -b -i test.prm -e results/err.txt
+        cd -
+    - name: Check the test results
+      shell: bash
```
popescu-v, line 66: Why `bash`? Is it for the Windows case (where default shell is different)? If so, then I would set the shell to `bash` only for steps which are not Linux-specific.

packaging/install.cmake
```diff
@@ -67,10 +67,20 @@ endif()
...
     set(KHIOPS_MPI_EXTRA_FLAG "--allow-run-as-root --quiet")
+    set(ADDITIONAL_EN_VAR "export OMPI_MCA_btl_vader_single_copy_mechanism=none # issue on docker")
+    if(IS_FEDORA_LIKE)
+      set(ADDITIONAL_EN_VAR "${ADDITIONAL_EN_VAR}\nexport PSM3_DEVICES=self # issue one rocky linux")
+    endif()
   endif()
+
+  # Add header comment to the variable definition (if any variable is defined)
+  if(ADDITIONAL_EN_VAR)
+    set(ADDITIONAL_EN_VAR "# Additional variables for MPI\n${ADDITIONAL_EN_VAR}")
```
popescu-v, line 81: s/for MPI/for OpenMPI/ ?

.github/actions/test-khiops-on-iris/action.yml
bruno-at-orange: You're right, on standard environment, it is not enough. But we're in a controlled environment.

packaging/install.cmake
bruno-at-orange: No, it is generic. This line may be generated for openmpi or mpich

popescu-v (changes requested):
There are several comments pending.

.github/actions/test-khiops-on-iris/action.yml
bruno-at-orange: because we use this action on windows

.github/actions/test-khiops-install/action.yml
bruno-at-orange: The shell definition is mandatory in actions

.github/actions/test-khiops-install/action.yml
popescu-v: But it is possible at the step level AFAIK.

.github/actions/test-khiops-install/action.yml
popescu-v: I see: for the OS-agnostic and Windows-specific steps we should indeed keep it IMHO, right.

.github/actions/test-khiops-on-iris/action.yml
popescu-v: Here, I think we can remove `shell: bash` as we are on Linux and the default `sh` shell should work AFAIU.

.github/actions/test-khiops-on-iris/action.yml
popescu-v: Yes, Ok. Closing the thread.

packaging/install.cmake
popescu-v: OK, fair enough.

.github/actions/test-khiops-install/action.yml
popescu-v: Yes, right.

.github/actions/test-khiops-on-iris/action.yml
popescu-v: Ok, we need to set the shell for actions, right.

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,57 @@
 ---
 name: Test Khiops Installation
```
popescu-v, line 2: Perhaps we can specify the "default" shell once and for all? Something like (see also https://github.com/actions/runner/blob/main/docs/adrs/0549-composite-run-steps.md#shell-and-working-directory):

```
inputs:
  default_shell:
    description: 'Bash'
    default: bash
```

And then in each step:

```
shell: ${{ inputs.default_shell }}
```

Thus, we set `bash` only once and we refer to it in each step?

N.B.: It is true that this would turn the shell into a "parameter" of the action; it could be overridden in workflows, via 
```
with: 
  default_shell: sh
```
for example. But I do not see an issue with this (we are not obliged to customize the shell from the workflow). The advantage I see in this approach is that we only set the "default" shell at one place; and just refer to it in each step.

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,57 @@
...
-    - name: Python setup for Linux
-      if: runner.os == 'Linux'
+    - name: Print status for debugging
       shell: bash
+      # We force the process number to 4 because the runner has only 2 proc.
+      # With 2 proc khiops-env switches to serial and mpi is not used.
+      # The env var OMPI_MCA_rmaps_base_oversubscribe is for openmpi it corresponds to
+      # the flag --oversubscribe (Nodes are allowed to be oversubscribed)
+      env:
+        KHIOPS_PROC_NUMBER: '4'
```
popescu-v: You don't need to quote the values here AFAIK.

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,57 @@
...
+        OMPI_MCA_rmaps_base_oversubscribe: 'true'
```
popescu-v: You can just leave `true` as a value (without quotes).

.github/actions/test-khiops-install/action.yml
```diff
@@ -1,74 +1,57 @@
...
+        if [ $LINE_NUMBER -gt 1 ] ;
+        then 
+          echo "::error::Unexpected output in khiops_coclustering scripts"
+          false
+        fi
+    - name: Check process number
       shell: bash
+      env:
+        KHIOPS_PROC_NUMBER: '4'
+        OMPI_MCA_rmaps_base_oversubscribe: 'true'
```
popescu-v: The values of the env vars can be left unquoted (as this is YAML).

.github/actions/test-khiops-on-iris/action.yml
```diff
@@ -0,0 +1,78 @@
...
+          echo "KHIOPS_CC_SCRIPT=$KHIOPS_CC_SCRIPT" >> "$GITHUB_ENV"
+        else
+          echo "KHIOPS_SCRIPT=khiops" >> "$GITHUB_ENV"
+          echo "KHIOPS_CC_SCRIPT=khiops_coclustering" >> "$GITHUB_ENV"
+        fi
+    - name: Check Khiops installation
+      shell: bash
+      env:
+        KHIOPS_PROC_NUMBER: '4'
+        OMPI_MCA_rmaps_base_oversubscribe: 'true'
```
popescu-v: The values can be left unquoted (see previous comments on this topic above).

.github/actions/test-khiops-on-iris/action.yml
```diff
@@ -0,0 +1,78 @@
...
+      run: |
+        "$KHIOPS_SCRIPT" -s
+        "$KHIOPS_CC_SCRIPT" -s
+    - name: Run Khiops tests
+      shell: bash
+      env:
+        KHIOPS_PROC_NUMBER: '4'
+        OMPI_MCA_rmaps_base_oversubscribe: 'true'
```
popescu-v: The values can be left unquoted.

.github/actions/test-khiops-on-iris/action.yml
popescu-v: Ok, then could you add a comment before the `if`, something like `# Debian-based OS:`?

.github/actions/test-khiops-install/action.yml
bruno-at-orange: I first wrote without quotes but the pre-commit disagrees: string is expected

.github/actions/test-khiops-install/action.yml
bruno-at-orange: I think it adds complexity. For me it is not a problem to set the shell at each step...

.github/actions/test-khiops-install/action.yml
popescu-v: This is weird, as we have these w/o quotes on `khiops-python` and pre-commit passes: e.g. https://github.com/KhiopsML/khiops-python/blob/a1952b218a18696d3c9930867103581820c3e103/.github/workflows/pip.yml#L72.

.github/actions/test-khiops-install/action.yml
popescu-v: OK, fair enough.

.github/actions/test-khiops-install/action.yml
popescu-v: And, locally, on the `khiops` repo as well, `yamlfix` succeeds when these 2 env vars are unquoted.

popescu-v: Oh, I see, it's `check-jsonschema` that fails. Its version needs to be updated to 0.28.6 in `.pre-commit-config.yaml`. Thus, pre-commit will succeed with these 2 environment variables without quotes (which is valid YAML).

.github/actions/test-khiops-on-iris/action.yml
bruno-at-orange: done

.github/actions/test-khiops-on-iris/action.yml
bruno-at-orange: done

.github/actions/test-khiops-on-iris/action.yml
bruno-at-orange: done

.github/actions/test-khiops-install/action.yml
bruno-at-orange: done

.github/actions/test-khiops-install/action.yml
bruno-at-orange: done

.github/workflows/pack-debian.yml
```diff
@@ -108,22 +108,16 @@ jobs:
...
       - name: Check Khiops installation
-        run: |
-          khiops-env --env
-          khiops -s
-          khiops_coclustering -s
-      - name: Test Khiops installation
-        continue-on-error: ${{ matrix.os == 'debian:11' }}
         uses: ./.github/actions/test-khiops-install
+      - name: Test Khiops on Iris dataset
+        continue-on-error: ${{ matrix.os == 'debian:11' }}
```
popescu-v: Why do we need this condition?

bruno-at-orange: I think there was a problem with debian 11 on  last release. I remove this

bruno-at-orange: done

popescu-v (approved):
LGTM. Can be merged as soon as the CI succeeds, IMHO.

PR #321, opened 2024-06-27
Creation of the results directory in kht_test (#320)

bruno-at-orange:
* Creation of the results directory in kht_test

If the results directory is missing, we create it before launching the binary to be tested. Because, when MODL fails before writing any file, it doesn't create the results dir. In that case, kht_test fails to write the log files.

* remove unnecessary os.getcwd()


marcboulle (approved):
LGTM

PR #325, opened 2024-07-01
307 cherry pick

bruno-at-orange:
cherry-picks from dev-v10 for #313


bruno-at-orange:
mixed up with cherry-picks

PR #326, opened 2024-07-01
Drop macos-11 support, which is phased out by GitHub

### LINKED ISSUE #308: Drop MacOS 11 support
#### Description
Indeed, GitHub runners are also dropping support for the `macos-11` runners, which are deprecated and will be removed on June 28th, 2024: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners

#### Questions/Ideas
- Addressing this issue will consist in removing the `macos-11` runner / target in all CI workflows.
- This should be done rapidly, otherwise some of the CI workflows would stop being green.

popescu-v:
closes #308


bruno-at-orange (approved):
LGTM

PR #330, opened 2024-07-03
Enable executable and lib signatures in macOS

folmos-at-orange:
- Add to the CI conda workflow the necessary environment variables so the macOS packages are signed (only on tag). The necessary variables are all set to values from GitHub organization secrets.
- Add signature for the KNI library

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -53,19 +53,27 @@ jobs:
         run: |
           wget https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX10.10.sdk.tar.xz
           sudo tar -zxvf MacOSX10.10.sdk.tar.xz -C /opt
-      - name: Put the Khiops Conda version in the environment
+      - name: Set the environment variables
         shell: bash
         run: |
-          # Obtain the Khiops version
+          # Put the Khiops package version
+          # The conda version cannot have '-' as a character so we eliminate it
           if [[ "${{ github.ref_type }}" == "tag" ]]
           then
             KHIOPS_RAW_VERSION="${{ github.ref_name }}"
           else
             KHIOPS_RAW_VERSION="$(./scripts/khiops-version)"
           fi
-
-          # The conda version cannot have '-' as a character so we eliminate it
           echo "KHIOPS_VERSION=$(echo $KHIOPS_RAW_VERSION | sed 's/-//')" >> "$GITHUB_ENV"
+
+          # On tag and macOS: Set the environment variables to sign the binaries
+          if [[ "${{ runner.os }}" == "macOS" && "${{ github.ref_type }}" == "tag" ]]
+          then
+            echo "KHIOPS_APPLE_CERTIFICATE_COMMON_NAME=${{ secrets.KHIOPS_APPLE_CERTIFICATE_COMMON_NAME }}" >> "$GITHUB_ENV"
+            echo "KHIOPS_APPLE_CERTIFICATE_BASE64=${{ secrets.KHIOPS_APPLE_CERTIFICATE_BASE64 }}" >> "$GITHUB_ENV"
+            echo "KHIOPS_APPLE_CERTIFICATE_PASSWORD=${{ secrets.KHIOPS_APPLE_CERTIFICATE_PASSWORD }}" >> "$GITHUB_ENV"
+            echo "KHIOPS_APPLE_TMP_KEYCHAIN_PASSWORD=${{ secrets.KHIOPS_APPLE_TMP_KEYCHAIN_PASSWORD }}" >> "$GITHUB_ENV"
+          fi
       - name: Build conda packages (Windows)
         if: runner.os == 'Windows'
         run: conda build --output-folder ./build/conda ./packaging/conda
--- a/packaging/conda/build.sh
+++ b/packaging/conda/build.sh
@@ -15,7 +15,8 @@ fi
 cmake --fresh --preset $CMAKE_PRESET -DBUILD_JARS=OFF -DTESTING=OFF
 
 # Build MODL and MODL_Coclustering
-cmake --build --preset $CMAKE_PRESET --parallel --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer 
+cmake --build --preset $CMAKE_PRESET --parallel \
+  --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer
 
 # Move the binaries to the Conda PREFIX path
 mv ./build/$CMAKE_PRESET/bin/MODL* "$PREFIX/bin"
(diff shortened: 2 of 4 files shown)


.github/workflows/conda.yml
```diff
@@ -53,19 +53,27 @@ jobs:
...
-          # The conda version cannot have '-' as a character so we eliminate it
           echo "KHIOPS_VERSION=$(echo $KHIOPS_RAW_VERSION | sed 's/-//')" >> "$GITHUB_ENV"
+
+          # On tag and macOS: Set the environment variables to sign the binaries
+          if [[ "${{ runner.os }}" == "macOS" && "${{ github.ref_type }}" == "tag" ]]
+          then
+            echo "KHIOPS_APPLE_CERTIFICATE_COMMON_NAME=${{ secrets.KHIOPS_APPLE_CERTIFICATE_COMMON_NAME }}" >> "$GITHUB_ENV"
+            echo "KHIOPS_APPLE_CERTIFICATE_BASE64=${{ secrets.KHIOPS_APPLE_CERTIFICATE_BASE64 }}" >> "$GITHUB_ENV"
+            echo "KHIOPS_APPLE_CERTIFICATE_PASSWORD=${{ secrets.KHIOPS_APPLE_CERTIFICATE_PASSWORD }}" >> "$GITHUB_ENV"
+            echo "KHIOPS_APPLE_TMP_KEYCHAIN_PASSWORD=${{ secrets.KHIOPS_APPLE_TMP_KEYCHAIN_PASSWORD }}" >> "$GITHUB_ENV"
```
popescu-v, line 75: I would have done this directly in an `env:` section of the steps where these secrets are needed:
```
env:
  KHIOPS_APPLE_CERTIFICATE_COMMON_NAME: ${{ secrets.KHIOPS_APPLE_CERTIFICATE_COMMON_NAME }}
```

etc, as these secrets are not _computed_ in this step, they come from GitHub settings; so, we can add them to the `env:` as we whish; see also e.g. https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#example-using-bash

packaging/conda/build.sh
```diff
@@ -15,7 +15,8 @@ fi
 cmake --fresh --preset $CMAKE_PRESET -DBUILD_JARS=OFF -DTESTING=OFF
 
 # Build MODL and MODL_Coclustering
-cmake --build --preset $CMAKE_PRESET --parallel --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer 
+cmake --build --preset $CMAKE_PRESET --parallel \
+  --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer
```
popescu-v, line 19: This change does not seem related to the current PR; was it required by the pre-commit hooks?

.github/workflows/conda.yml
folmos-at-orange: I want to add them just when they are necessary (macOS + tag). Before I used an `env:` in the proper step, but the necessary `${{ ...}}` statements were very cryptic. This is clearer and easier to debug.

**EDIT*** In addition If the variable `KHIOPS_CERTIFICATE_COMMON_NAME` is defined it triggers the signature. There is no way to "not define" a variable in an `env:` statement.

packaging/conda/build.sh
folmos-at-orange: No, just a small editorial change because the line was too long.

packaging/conda/build.sh
```diff
@@ -101,11 +103,16 @@ then
   # identity and makes the build fail!
   CODESIGN="/usr/bin/codesign"
 
-  # Sign the MODL executable and check
+  # Obtain the path of the real KNI (not the symlinks)
+  KNI_PATH=$(file --mime --no-dereference $PREFIX/lib/libKhiops* | grep x-mach-binary | cut -d: -f1)
```
popescu-v: I'm not sure I understand this well; some DLLs give us `x-sharedlib`. What do we want to achieve here?

folmos-at-orange: In macOS this is the output of the command is
```bash
(kh-dev) folmos@MC-JWMHQQ2JYL:~/work/khiops/khiops (add-certs-to-conda)
(11:59:36)> file -I --no-dereference  ~/Downloads/khiops-conda-osx-arm64-4/osx-arm64/kni-10.3.0a.2-h2833df3_0/lib/libKhiopsNativeInterface.*
/Users/folmos/Downloads/khiops-conda-osx-arm64-4/osx-arm64/kni-10.3.0a.2-h2833df3_0/lib/libKhiopsNativeInterface.10.2.2-b.2.dylib: application/x-mach-binary; charset=binary
/Users/folmos/Downloads/khiops-conda-osx-arm64-4/osx-arm64/kni-10.3.0a.2-h2833df3_0/lib/libKhiopsNativeInterface.10.dylib:         inode/symlink; charset=binary
/Users/folmos/Downloads/khiops-conda-osx-arm64-4/osx-arm64/kni-10.3.0a.2-h2833df3_0/lib/libKhiopsNativeInterface.dylib:            inode/symlink; charset=binary
```
So the mime type of dynamic libraries in macOS is `application/x-mach-binary`

.github/workflows/conda.yml
popescu-v: Yeah, I see. Moreover, there is also no way to unset the env vars from `GITHUB_ENV` either...

packaging/conda/build.sh
popescu-v: Ok.

packaging/conda/build.sh
popescu-v: Perhaps `readlink -f` might help achieve the goal here?

folmos-at-orange: I could change to `KNI_PATH=$(readlink -f $PREFIX/lib/libKhiops* | uniq)`

popescu-v: Yeah, it seems more readable and straightforward IMHO.

popescu-v: Perhaps add ` | head -n 1` to ensure that structurally there is a unique entry?

folmos-at-orange: Ok & Done.

popescu-v: I don't see  the ` | head -n 1`...

folmos-at-orange: It should be one, if not, there is a problem and better not to hide it.

popescu-v: Oh, ok then the build would eventually fail because of this, if the KNI path is not unique.

popescu-v (commented):
Rebase current branch on `dev-v10`. Then cherry-pick on `dev` as well.

popescu-v (approved):
LGTM

PR #340, opened 2024-07-30
171 add build for ubuntu 2404

### LINKED ISSUE #171: Add build for ubuntu 24.04
#### Description
Ubuntu 24.04 will be latest LTS in a month. We should add it to the CI/CD process to start testing it. 

#### Questions/Ideas
- Add Ubuntu 24.04 to `packaging/dockerfiles`
  - The image is already available in docker hub https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
- Add Ubuntu 24.04 to `.github/workflows/pack-deb.yml`
- Demand synchronization in `LearningTest` infra
- Maybe eliminate 18.04 from `packaging/dockerfiles` ?

bruno-at-orange:
- Add new docker image for 24.04
- Build new deb packages for 24.04
- Add a test on 24.04 for conda packages

We have to test the new deb packages on LearningTest environment (new issue on [gitlab](https://gitlab.tech.orange/khiops/learningtest/runner/-/issues/18))

### DIFF
--- a/packaging/dockerfiles/Dockerfile.ubuntu24.04
+++ b/packaging/dockerfiles/Dockerfile.ubuntu24.04
@@ -1,0 +1,25 @@
+FROM ubuntu:24.04
+LABEL maintainer="khiops.team@orange.com"
+LABEL description="Container with the dependencies to build and package Khiops"
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN true \
+    && apt-get -y update \
+    && apt-get -y --no-install-recommends install \
+    build-essential \
+    cmake \
+    coreutils \
+    debhelper \
+    openjdk-8-jdk \
+    devscripts \
+    fakeroot \
+    libmpich-dev \
+    openmpi-bin \
+    libopenmpi-dev \
+    nano \
+    ninja-build \
+    curl \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get clean \
+    && true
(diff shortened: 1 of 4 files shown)


packaging/dockerfiles/Dockerfile.ubuntu24.04
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:24.04
```
popescu-v, line 1: IMHO we could factor out the Dockerfiles per distribution: one for Ubuntu, one for Rocky; the version number could be parameterized from the GitHub workflows; see e.g. https://github.com/KhiopsML/khiops-python/blob/dev/packaging/docker/khiopspydev/Dockerfile.ubuntu for the Dockerfile and https://github.com/KhiopsML/khiops-python/blob/dev/.github/workflows/dev-docker.yml for their instantiations.

popescu-v (changes requested):
Technically, LGTM. However, there is a comment that could be addressed IMHO.

packaging/dockerfiles/Dockerfile.ubuntu24.04
bruno-at-orange: I disagree, it may be some specific packages on each distro. For instance, on ubuntu 20.04 we set a specific version of cmake.

popescu-v: OK, I see your point.

popescu-v (approved):
LGTM

PR #346, opened 2024-08-09
update khiops-env to display error (#284)

bruno-at-orange:
* Update khiops-env to display error on Linux

On Linux, the khiops-env script doesn't print anything in the stdout/err anymore (it is already the case on windows). It only populates environment variables if possible. In case of errors (java or mpi) it populates 2 new variables KHIOPS_MPI_ERROR and KHIOPS_JAVA_ERROR

The khiops script use notify-send to diplay errors to users. It implies a new dependency in the khiops package: libnotify on rocky and libnotify-bin on debian. Errors are also logs into /var/log/syslog by using the 'logger' command.

cherry-pick from dev


popescu-v (approved):
LGTM (cherry-pick of https://github.com/KhiopsML/khiops/pull/284).

PR #363, opened 2024-08-30
150 extend the copyright script for python files

### LINKED ISSUE #150: Extend the copyright script for python files
#### Description
Currently the script `scripts/update_copyright.py` only changes `.cpp` and `.h` files. It should also consider python and shell files

#### Questions/Ideas
- The comment operator depends on the extension:
  - `.py`, `.sh` : `#`
  - `.cmd`, `.bat` : `REM`
  - `.h`, `.cpp`: `//`

folmos-at-orange:


.cmake-format.py
```diff
@@ -1,3 +1,7 @@
+# Copyright (c) 2024 Orange. All rights reserved.
```
popescu-v, line 1: s/bannor/banner/ in commit message for https://github.com/KhiopsML/khiops/pull/363/commits/5cdbba9b0097166236412081701004e7685ed22c

scripts/update-copyright.py
```diff
@@ -1,44 +1,74 @@
...
 def update_copyright(file_path):
-    """Updates the copyright notice of a file"""
-    print(f"Updating {file_path}")
+    """Updates the copyright notice of a file if necessary"""
+
+    # Obtain the comment prefix string from the file extension
+    _, ext = os.path.splitext(file_path)
+    if ext in (".h", ".c", ".hpp", ".cpp", ".java"):
+        comment_prefix = b"// "
+    elif ext == ".py":
```
popescu-v, line 53: Perhaps we should do the same for Bash scripts as well (`khiops-env`, `khiops`, `khiops_coclustering`) and their Windows counterparts?

test/MemoryStatsVisualizer/collect_stats_per_task.py
```diff
@@ -1,3 +1,7 @@
+# Copyright (c) 2024 Orange. All rights reserved.
+# This software is distributed under the BSD 3-Clause-clear License, the text of which is available
+# at https://spdx.org/licenses/BSD-3-Clause-Clear.html or see the "LICENSE" file for more details.
+
 # -*- coding: utf-8 -*-
```
popescu-v: AFAIK, this special comment should be the first line in the file, if we keep it. But for Python 3, this is not necessary anymore AFAIK. So: either remove it, or add special logic to keep at the top of the file.

The same applies to all files that are concerned.

popescu-v (changes requested):
See comments.

scripts/update-copyright.py
folmos-at-orange: The issue stated that we'll do it only for `.py`. It was discussed in a weekly.

test/MemoryStatsVisualizer/collect_stats_per_task.py
folmos-at-orange: Ok, it seems that the standard encoding is UTF8 so i'll remove these.

scripts/update-copyright.py
popescu-v: Ok, I understand your point. However the issue (#150) states "Extend the copyright script for python and script files". So, I would edit the issue description according to what was convened (I can see the comment to the issue).

test/MemoryStatsVisualizer/collect_stats_per_task.py
popescu-v: Ok, that special comment was only for Python 2 (standard encoding was ASCII AFAIR); not needed for Python 3.

scripts/update-copyright.py
folmos-at-orange: Done

test/MemoryStatsVisualizer/collect_stats_per_task.py
folmos-at-orange: Done

popescu-v (approved):

PR #364, opened 2024-08-30
Remove setup-ninja action

### LINKED ISSUE #102: Remove unmaintained setup-ninja action
#### Description
The [setup-ninja action](https://github.com/marketplace/actions/setup-ninja) is unmaintained. Given that it just installs ninja we should manage ourselves its setup in CI workflows.
#### Questions/Ideas
- In windows it is already installed
- In linux ?
- In mac ?

folmos-at-orange:


.github/actions/build-khiops/action.yml
```diff
@@ -29,9 +29,16 @@ runs:
...
         sudo update-alternatives --set mpi /usr/bin/mpicc.openmpi
         sudo update-alternatives --set mpirun /usr/bin/mpirun.openmpi
-    - name: Setup Ninja
-      uses: ashutoshvarma/setup-ninja@v1.1
-    - name: Load Visual C++ Environment Variables (Windows only)
+    - name: Setup Ninja (Linux)
+      if: runner.os == 'Linux'
+      shell: bash
+      run: sudo apt-get install ninja-build
+    - name: Setup Ninja (macOS)
```
popescu-v, line 36: We do not build native packages on Mac OS, do we? And for Conda, this is managed by Conda itself, right? So, I would remove the MacOS part for now.

popescu-v (commented):
one comment

.github/actions/build-khiops/action.yml
folmos-at-orange: Yes, but there is an action which depend on this. Better let this there so when fix the mac package is done (it cost nothing).

popescu-v: Ok to stay isofunctional to what we had before.

popescu-v (approved):

PR #367, opened 2024-09-03
100 implement tagversion coherence in ci

### LINKED ISSUE #100: Implement tag/version coherence in CI
#### Description
_Re-specified after meeting 03/04/2024_

#### Questions/Ideas
- Eliminate `scripts/khiops-package-version`
- Implement in windows `scripts/khiops-version.bat`
- Try to make it work in CMake
- Implement action `check-tag-version`
  - On `tag` event it checks that the source version is the same as the tag.
  - Can we rollback the tag automatically on failure ?

#### Description
PR #74 implemented helper scripts related to the Khiops version:
- `scripts/khiops-version`: obtains the Khiops version string from the sources
- `scripts/khiops-package-version`: create a version string by examining
  - The output of `scripts/khiops-version`
  - A Git "ref_name", typically a tag or a SHA
 
The objective of this scripts is to ease CI/CD tasks
- `scripts/khiops-version`: A clean single way to obtain the source version, as oppossed to use `sed` incantations in the middle CI/CD scripts. The script is itself a `sed` incantation but it is centralized and abstracted away (we may change the implementation as we wish).
- `scripts/khiops-package-version`: A version generator that can be used in CI/CD task on _any_ commit or tagged. When it is a tag and it matches the source version it will just output the source version. Otherwise it will output a `preview` version with information to retrieve the commit where it was generated. This allows at the same time to automate the package naming for releases (tag push) or quick tests (manual generation).
- 
#### Questions/Ideas
- Generalize the application of this script to the linux packaging
   - We can discuss the naming and choices of the `khiops-package-version` script
- The replacement in the main `CMakeLists.txt` file is
```cmake
execute_process(
  COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/scripts/khiops-version" "HEAD" 
  OUTPUT_VARIABLE VERSION 
  ERROR_QUIET 
  OUTPUT_STRIP_TRAILING_WHITESPACE
)
```
- However `khiops-package-version` doesn't work because it doesn't comply with cmake restrictions
  - There have been [requests to relax the version parsing](https://gitlab.kitware.com/cmake/cmake/-/issues/23649) without response (6 years)
- We could either:
  - Use our custom package version in the package cmake scripts
  - Change the behavior of `khiosp-package-version` to comply with CMake versioning scheme.

folmos-at-orange:


.github/workflows/pack-debian.yml
```diff
@@ -191,6 +192,9 @@ jobs:
         with:
           allowUpdates: true
           artifacts: '*.deb'
+          body: |
+            **This release is for testing purporses only and there is no support for it.**
```
popescu-v: s/purporses/purposes/

.github/workflows/conda.yml
```diff
@@ -135,6 +135,10 @@ jobs:
     permissions:
       contents: write
     steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+      - name: Check the tag consistency with the source version
```
popescu-v: As this is a build -> test[ -> release] workflow, why not check the tag consistency with the version from the outset,  _if the workflow is launched on a tag_, i.e. in the `build` job? Thus, we would not waste time with the build -> test jobs if the workflow is launched on a tag, because we would need to re-launch the whole workflow anyway, right?

.github/workflows/pack-debian.yml
```diff
@@ -166,6 +165,8 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
+      - name: Check the tag consistency with the source version
```
popescu-v: Check this in the `build` job if the workflow is launched on a Git tag (see above)?

.github/workflows/pack-nsis.yml
```diff
@@ -155,6 +140,10 @@ jobs:
     permissions:
       contents: write
     steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+      - name: Check the tag consistency with the source version
```
popescu-v: Check this in the `build` job if the workflow is launched on a Git tag (see above)?

.github/workflows/pack-nsis.yml
```diff
@@ -180,7 +169,9 @@ jobs:
         with:
           allowUpdates: true
           artifacts: khiops-*-setup.exe,KNI-*.zip,kni-transfer-*.zip
-          body: '**For testing purposes only**'
+          body: |
+            **This release is for testing purporses only and there is no support for it.**
```
popescu-v: s/purporses/purposes/

.github/workflows/pack-rpm.yml
```diff
@@ -157,6 +157,8 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
+      - name: Check the tag consistency with the source version
```
popescu-v: Check this in the `build` job if the workflow is launched on a Git tag (see above)?

.github/workflows/pack-rpm.yml
```diff
@@ -182,7 +184,9 @@ jobs:
         with:
           allowUpdates: true
           artifacts: '*.rpm'
-          body: '**For testing purposes only**'
+          body: |
+            **This release is for testing purporses only and there is no support for it.**
```
popescu-v: s/purporses/purposes/

popescu-v (commented):
See comments.

General highlight: I would suggest performing this check at the `build` stage when the workflows are launched on Git tags.

.github/workflows/conda.yml
folmos-at-orange: Yes, indeed, i'm working on a fix. The build job seems the good place to check.

folmos-at-orange: Done

.github/workflows/pack-debian.yml
folmos-at-orange: Done as above.

.github/workflows/pack-nsis.yml
folmos-at-orange: Ok & Done as above

.github/workflows/pack-nsis.yml
folmos-at-orange: Ok & Done

.github/workflows/pack-rpm.yml
folmos-at-orange: Ok & Done as above

.github/workflows/pack-rpm.yml
folmos-at-orange: Ok & Done

.github/workflows/pack-debian.yml
folmos-at-orange: Ok & Done

.github/workflows/pack-debian.yml
```diff
@@ -37,13 +36,16 @@ jobs:
...
+        if: github.event_name == 'push'
```
popescu-v: Why not rather `if: github.ref_type == 'tag'`? This is what I had in mind... so that this check is also performed when the workflow is launched manually, on a specified Git tag.

Thus, this check would be done in two cases:
- when we push _on a tag_ (`push: tags: '*'`)
- when we manually launch the workflow _on a tag_ (`workflow_dispatch`).

folmos-at-orange: Ok & Done.

.github/actions/check-tag-version/action.yml
```diff
@@ -0,0 +1,17 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Check that the Khiops source version matches the input tag
+      shell: bash
+      run: |
+        SOURCE_VERSION=$(./scripts/khiops-version)
+        if [[ "$SOURCE_VERSION" != "${{ github.ref_name }}" ]]
+        then
+          echo "::error::Incompatible source version '$SOURCE_VERSION' with tag '${{ github.ref_name }}'. Modify it in src/Learning/KWUtils/KWKhiopsVersion.h."
```
popescu-v: The "it" in "Modify it" refers to the Khiops version. However, in order for this to be clear, IMHO we need to rephrase this a little bit: either "Modify source version in src/..", or "Incompatible tag ... with source version... Modify it in src/..."

folmos-at-orange: Ok I modified it.

popescu-v (approved):
LGTM

PR #368, opened 2024-09-04
100 implement tagversion coherence in ci (backport dev-v10)

### LINKED ISSUE #100: Implement tag/version coherence in CI

folmos-at-orange:


popescu-v (approved):
LGTM

PR #395, opened 2024-10-07
392 use absolute path to mpiexec in the unix khiops env scripts for native installations

popescu-v:


bruno-at-orange (approved):
LGTM

bruno-at-orange:
need to be ported to v11

PR #397, opened 2024-10-07
Use absolute path to `mpiexec` in `khiops_env` on all UNIX installations

### LINKED ISSUE #392: Use Absolute Path to `mpiexec` in the UNIX `khiops_env` Scripts for Native Installations
#### Description
Currently, the UNIX `khiops_env` script for native installations sets unqualified `mpiexec` in the `KHIOPS_MPI_COMMAND` environment variable.

In Rocky Linux systems, this relies on setting the appropriate `mpi` environment module, which `khiops_env` does. However, the scope of the effects of environment module load are limited to the current (interactive) shell session. Hence, the approach fails when `khiops_env` is used from the Khiops Python library where *separate* subprocesses are used for:
- calling `khiops_env --env`, in order to retrieve the environment variables and populate Python process's environment
- calling Khiops binaries; for these subsequent subprocesses, `mpiexec` is not in the `PATH`.

#### Questions/Ideas
- A solution is to systematically use the absolute path to `mpiexec` when `KHIOPS_MPI_COMMAND` is initialized
  - this can be achieved, for all UNIXes, thus:
    `KHIOPS_MPI_COMMAND="$(type -P mpiexec) ..."` 
- Thus, CMAKE logic around `@KHIOPS_MPI_PATH@` could be discarded

### Context
- Khiops version
- Log file (use `khiops -e log.txt`): 10.2.3-b.2
- Scenario file (use `khiops -o scenario._kh`): any
- OS description (use `khiops -s`): UNIXes (mostly Linux native; but Conda would also be addressed through the proposed approach)

popescu-v:
closes #392

### DIFF
--- a/packaging/install.cmake
+++ b/packaging/install.cmake
@@ -72,7 +72,6 @@ if(UNIX)
     set(KHIOPS_PATH "$(get_script_dir)")
     set(KHIOPS_COCLUSTERING_PATH "$(get_script_dir)")
     set(GET_PROC_NUMBER_PATH "$(get_script_dir)")
-    set(KHIOPS_MPI_PATH "$(get_script_dir)") # We use mpiexec in conda's bin directory
     set(IS_CONDA_VAR "\n# Inside conda environment\nexport _IS_CONDA=true")
   else()
     if(IS_FEDORA_LIKE)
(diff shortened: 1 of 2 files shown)


bruno-at-orange (approved):
LGTM

PR #400, opened 2024-10-11
Run MODL in parallel in run-standard-tests (fix)

bruno-at-orange:
KhiopsMPIProcessNumber is deprecated in kht_test.
We use the new API "-p 4" to run the tests in parallel.

PR #402, opened 2024-10-11
Run MODL in parallel in run-standard-tests (fix) (#400)

bruno-at-orange:
KhiopsMPIProcessNumber is deprecated in kht_test.
We use the new API "-p 4" to run the tests in parallel.


popescu-v (approved):
LGTM

PR #407, opened 2024-10-11
405 cmake issue when building conda on fedora

### LINKED ISSUE #403: The conda build uses the wrong compiler
#### Description
On conda gcc/g++ compiler are not named gcc or g++ but `x86_64-conda-linux-gnu-c++`. When we use the cmake presets, cmake point to the system wide compiler (as it is required in the preset) and not the conda compiler.  Compiler packages set environment variables to point cmake to the [correct compiler](https://github.com/ContinuumIO/anaconda-issues/issues/10910#issuecomment-491614100).

#### Questions/Ideas
-  do not use presets and let cmake find the conda compiler

### LINKED ISSUE #405: CMake issue when building conda on fedora
#### Description
When we configure the project on fedora. A fatal error is triggered if the variable MPI_BIN is not defined. On conda, this variable is not required. As it stands, we can not build conda package on fedora distro.
It is related to #391

#### Questions/Ideas
- trigger the error only on conda environment

### Context
- OS description (use `khiops -s`) conda / fedora

bruno-at-orange:
This PR contains fixes and improvements that I have made to package khiops on conda-forge. 
There is a major fix concerning the use of the compiler of conda instead of the system wide one (commit 2b317f4ba9217a9d2ba92cd51bc6b54e2cce83df)

close #403 
close #405

### DIFF
--- a/packaging/conda/README.md
+++ b/packaging/conda/README.md
@@ -12,12 +12,7 @@ You'll need `conda-build` installed in your system.
 export KHIOPS_RAW_VERSION=$(scripts/khiops-version)
 export KHIOPS_VERSION=$(echo $KHIOPS_RAW_VERSION | sed 's/-//')
 
-# Windows
 conda build --output-folder ./khiops-conda packaging/conda
-
-# Linux/macOS
-# Note: We need the conda-forge channel to obtain the pinned versions of MPICH
-conda build --channel conda-forge --output-folder ./khiops-conda packaging/conda
 ```
 
 ### Signing the Executables in macOS
--- a/packaging/conda/README.md
+++ b/packaging/conda/README.md
@@ -1,16 +1,33 @@
 # Khiops Conda Packaging Scripts
 
 ## How to Build
-You'll need `conda-build` installed in your system.
+We need `conda-build` installed in the system.
+
+We need to make sure that Conda is configured to use conda-forge as its default channel and that the vanilla default channel (defaults) is removed, e.g. by writing:
+
+```bash
+$ conda config --add channels conda-forge
+$ conda config --remove channels defaults
+```
+
+Or if we want to keep the vanilla defaults channel, we could give the priority to conda-forge:
+
+```bash
+$ conda config --add channels conda-forge
+$ conda config --set channel_priority strict
+```
+
+Thus, the user's $HOME/.condarc file would be updated accordingly and --channel conda-forge would no longer be needed.
+
+In the CI this is ensured by the usage of miniforge-version in the invocation of the setup-miniconda action.
+
+To build `khiops-core` package, we need to run these commands at the root of the repo (it will leave a ready to use conda channel in `./khiops-conda`):
 
 ```bash
 # At the root of the repo
-# These commands will leave a ready to use conda channel in `./khiops-conda`
 
 # khiops version must be set before launching the build (can be retrieved with the script scripts/khiops-version).
-# The dash must be removed from the version to be compliant with the conda versionning policy
-export KHIOPS_RAW_VERSION=$(scripts/khiops-version)
-export KHIOPS_VERSION=$(echo $KHIOPS_RAW_VERSION | sed 's/-//')
+export KHIOPS_VERSION=$(scripts/khiops-version)
 
 conda build --output-folder ./khiops-conda packaging/conda
 ```
--- a/packaging/conda/build.sh
+++ b/packaging/conda/build.sh
@@ -3,29 +3,21 @@
 # Set-up the shell to behave more like a general-purpose programming language
 set -euo pipefail
 
-# Choose the build preset for macOS/Linux
-if [[ "$(uname)" == "Linux" ]]
-then
-  CMAKE_PRESET="linux-gcc-release"
-else
-  CMAKE_PRESET="macos-clang-release"
-fi
-
 # Configure project
-cmake --fresh --preset $CMAKE_PRESET -DBUILD_JARS=OFF -DTESTING=OFF
+cmake -B build/conda -S . -D BUILD_JARS=OFF -D TESTING=OFF -D CMAKE_BUILD_TYPE=Release -G Ninja
 
 # Build MODL and MODL_Coclustering
-cmake --build --preset $CMAKE_PRESET --parallel \
+cmake --build build/conda --parallel \
   --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer _khiopsgetprocnumber
 
 # Move the binaries to the Conda PREFIX path
-mv ./build/$CMAKE_PRESET/bin/MODL* "$PREFIX/bin"
-mv ./build/$CMAKE_PRESET/bin/_khiopsgetprocnumber* "$PREFIX/bin"
-mv ./build/$CMAKE_PRESET/bin/KNITransfer* "$PREFIX/bin"
-mv ./build/$CMAKE_PRESET/lib/libKhiopsNativeInterface* "$PREFIX/lib"
+mv ./build/conda/bin/MODL* "$PREFIX/bin"
+mv ./build/conda/bin/_khiopsgetprocnumber* "$PREFIX/bin"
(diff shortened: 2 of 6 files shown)


packaging/conda/README.md
```diff
@@ -12,12 +12,7 @@ You'll need `conda-build` installed in your system.
 export KHIOPS_RAW_VERSION=$(scripts/khiops-version)
 export KHIOPS_VERSION=$(echo $KHIOPS_RAW_VERSION | sed 's/-//')
 
-# Windows
 conda build --output-folder ./khiops-conda packaging/conda
-
-# Linux/macOS
-# Note: We need the conda-forge channel to obtain the pinned versions of MPICH
-conda build --channel conda-forge --output-folder ./khiops-conda packaging/conda
```
popescu-v, line 20: We need to make sure that Conda is configured to use `conda-forge` as its default channel and that the vanilla default channel (`defaults`) is removed, e.g. by writing:
```
$ conda config --add channels conda-forge
$ conda config --remove channels defaults
```
Or if we want to keep the vanilla `defaults` channel, we could give the priority to `conda-forge`:

```
$ conda config --add channels conda-forge
$ conda config --set channel_priority strict
```

Thus, the user's `$HOME/.condarc` file would be updated accordingly and `--channel conda-forge` would no longer be needed.

In the CI this is ensured by the usage of `miniforge-version` in the invocation of the `setup-miniconda` action.

packaging/conda/build.sh
```diff
@@ -91,8 +80,8 @@ then
     sudo security unlock-keychain -p "$KHIOPS_APPLE_TMP_KEYCHAIN_PASSWORD" kh-tmp.keychain
 
     # Add identity (certificate + private key) to the temporary keychain
-    echo "$KHIOPS_APPLE_CERTIFICATE_BASE64" \
-      | base64 --decode -i - -o kh-cert.p12
+    echo "$KHIOPS_APPLE_CERTIFICATE_BASE64" |
+      base64 --decode -i - -o kh-cert.p12
```
popescu-v: Do we need to have these on two separate lines?

packaging/conda/build.sh
```diff
@@ -3,29 +3,21 @@
...
-if [[ "$(uname)" == "Linux" ]]
-then
-  CMAKE_PRESET="linux-gcc-release"
-else
-  CMAKE_PRESET="macos-clang-release"
-fi
-
 # Configure project
-cmake --fresh --preset $CMAKE_PRESET -DBUILD_JARS=OFF -DTESTING=OFF
+cmake -B conda-build -S . -D BUILD_JARS=OFF -D TESTING=OFF -D CMAKE_BUILD_TYPE=Release -G Ninja
```
popescu-v: I would have kept the `build` directory name, to make the change minimal. Or, also `/build/conda/` instead of `/conda-build/` ?

packaging/conda/meta.yaml
```diff
@@ -1,6 +1,8 @@
+{% set version = os.environ.get('KHIOPS_VERSION') %}
+
 package:
   name: khiops-binaries
-  version: {{ os.environ.get('KHIOPS_VERSION') }}
+  version: {{ version.replace("-", ".") }}
```
popescu-v, line 5: I would write `{{ version.replace("-", "") }}` to remain consistent with the current behavior, and thus versioning scheme.

packaging/conda/meta.yaml
```diff
@@ -40,7 +42,7 @@ outputs:
 
   # khiops-core package
   - name: khiops-core
-    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    version: {{ version.replace("-", ".") }}
```
popescu-v, line 45: I would substitute "-" for "" to be consistent with current versioning scheme (see comment above).

packaging/conda/meta.yaml
```diff
@@ -61,13 +63,13 @@ outputs:
...
       commands:
-        - khiops_env --env 
+        - khiops_env --env
         - khiops -s
         - khiops_coclustering -s
 
   # kni package (do not need khiops-core as a runtime dependency)
   - name: kni
-    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    version: {{ version.replace("-", ".") }}
```
popescu-v, line 72: I would substitute "-" for "" to be consistent with current versioning scheme (see comment above).

packaging/conda/meta.yaml
```diff
@@ -81,7 +83,7 @@ outputs:
 
   # kni-transfer package (designed only to test kni)
   - name: kni-transfer
-    version: {{ os.environ.get('KHIOPS_VERSION') }}
+    version: {{ version.replace("-", ".") }}
```
popescu-v, line 86: I would substitute "-" for "" to be consistent with current versioning scheme (see comment above).

packaging/conda/meta.yaml
```diff
@@ -1,6 +1,8 @@
+{% set version = os.environ.get('KHIOPS_VERSION') %}
```
popescu-v: We can write directly:

```
{% set version = os.environ.get('KHIOPS_VERSION').replace("-", "") %}
```

packaging/conda/meta.yaml
popescu-v: With the modification in the comment above, we can write directly:
```
version: {{ version }}
```

for the `khiops-binaries` package as well as all outputs.

packaging/conda/meta.yaml
popescu-v: I would replace "-" with "" to remain consistent with the current versioning scheme.

popescu-v (changes requested):
Some comments, among which two changes in the Conda package versioning.

packaging/conda/README.md
bruno-at-orange: I added your comment in the README.md

packaging/conda/build.sh
bruno-at-orange: ok to `build/conda`

packaging/conda/build.sh
bruno-at-orange: no...

packaging/conda/meta.yaml
bruno-at-orange: done

packaging/conda/meta.yaml
bruno-at-orange: done

packaging/conda/build.sh
```diff
@@ -3,29 +3,21 @@
...
-cmake --build --preset $CMAKE_PRESET --parallel \
+cmake --build build-conda --parallel \
   --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer _khiopsgetprocnumber
 
 # Move the binaries to the Conda PREFIX path
-mv ./build/$CMAKE_PRESET/bin/MODL* "$PREFIX/bin"
-mv ./build/$CMAKE_PRESET/bin/_khiopsgetprocnumber* "$PREFIX/bin"
-mv ./build/$CMAKE_PRESET/bin/KNITransfer* "$PREFIX/bin"
-mv ./build/$CMAKE_PRESET/lib/libKhiopsNativeInterface* "$PREFIX/lib"
+mv ./build-conda/bin/MODL* "$PREFIX/bin"
```
popescu-v: What I had in mind was rather `./build/conda`; but `./build-conda` can also do, because there is no confusion with the `conda-build` / `conda build` command anymore.

packaging/conda/build.sh
popescu-v: Done.

packaging/conda/build.sh
popescu-v: Done with `build/conda`.

popescu-v:
Implemented the remaining agreed-upon modification suggestion in the comments: renaming of `build-conda`  to `build/conda`.

popescu-v (approved):

PR #414, opened 2024-10-18
383 bis khiops command return error 1 without any message

bruno-at-orange:
Backport of #393


popescu-v (approved):
LGTM (identical to PR #393 - already merged into `dev`, as expected).

PR #425, opened 2024-10-31
Let CMake choose the compiler in Conda on Windows as well

popescu-v:
Thus, there is no need to override target platform and generator toolset anymore.


bruno-at-orange (approved):
LGTM

PR #432, opened 2024-11-06
430 backport v11

### LINKED ISSUE #429: LearningTestTool Fails on Debian 10 Native Khiops Installations that Use OpenMPI
#### Description
Currently, in LearningTestTool, OpenMPI is instructed to run as root via the `OMPI_ALLOW_RUN_AS_ROOT` and `OMPI_ALLOW_RUN_AS_ROOT_CONFIRM` environment variables as hard-coded in the `/test/LearningTestTool/py/kht_test.py` script.

However, these variables are supported in OpenMPI since around 2018, via (at least) commit https://github.com/ax3l/ompi/commit/0cb8fc46c375f10328c6b1e403f8df9a5a591684.

But Debian 10 presumably contains an older OpenMPI version, which does not support these variables. It only supports the `--allow-run-as-root` command-line argument.

#### Questions/Ideas
- The fix consists in appending `--allow-run-as-root` to the `khiops_params` list in the `evaluate_tool_on_test_dir` function in `kht_test.py`.
  - This is a portable solution that would work for all supported platforms
- Note: this issue does _not_ impact the released packages, because they rely on the `khiops_env` script which uses the `--allow-run-as-root` command-line argument.

### Context
- Khiops version: 10.2.3-rc.1
- Log file (use `khiops -e log.txt`): N/A
- Scenario file (use `khiops -o scenario._kh`): Any
- OS description (use `khiops -s`): Debian 10

popescu-v:
backport of #430 on v11.

closes #429 on v11.

### DIFF
--- a/test/LearningTestTool/py/kht_test.py
+++ b/test/LearningTestTool/py/kht_test.py
@@ -341,7 +341,7 @@ def evaluate_tool_on_test_dir(
         os.environ[kht.KHIOPS_API_MODE] = "true"
 
         # Ajout de variables d'environements propres a OpenMPI, elles remplacent les parametres
-        # on peut ansi lancer indiferemment mpich ou openmpi
+        # on peut ainsi lancer indiferemment mpich ou openmpi
         if platform.system() == "Linux":
             # Supprime les traces en cas d'erreur fatale de khiops. Option --quiet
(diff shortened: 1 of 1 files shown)


marcboulle (approved):
LGTM

PR #435, opened 2024-11-07
433 backport v11

popescu-v:
Backport of #433 to v11, with two differences:

- Khiops version change is not included (as it is not applicable in the context of this v11 PR)
- WHATSNEW.txt gets 2 additions, the changelogs for 10.2.2 and 10.2.3, that were missing from the v11 "trunk".


marcboulle (approved):
LGTM

PR #521, opened 2025-01-10
495 khiops env env raises spurious error when environment modules are installed on non rpm linux operating systems

### LINKED ISSUE #495: `khiops_env --env` Raises Spurious Error When Environment Modules Are Installed on Non-RPM Linux Operating Systems
#### Description
Currently, `khiops_env` detects if environment module support is installed on the host Linux operating system. If so, then `khiops_env` attempts to unload the MPI environment module, and to load the appropriate MPI environment module.

On RPM-based Linux distributions (Rocky 8, 9), this works as expected. On vanilla Debian / Debian-based distributions, this also works as expected, because environment module support is not installed by default.

However, if, for some reason, environment module support is installed on the host Debian-like distribution (via `apt-get install environment-modules`), then `khiops_env` detects this support and assumes that the MPI environment module can be loaded, which is not true, because MPI is not packaged with environment module support on these Linux distributions.

This results in a spurious error which is propagated to stderr: `ERROR: Unable to locate a modulefile for 'mpi/openmpi-x86_64'`. This error has no effect on the functioning of the script, which exits with the `0` return code.

#### Questions/Ideas
- A solution is to redirect MPI environment module loading attempt's outputs (stdout and stderr) to `/dev/null`:
   ```bash
   module load mpi/openmpi-x86_64 &> /dev/null
   ```
### Context
- Khiops version: 10.2.4
- OS description (use `khiops -s`): Linux Debian and Ubuntu.

popescu-v:
closes #495


bruno-at-orange (approved):
LGTM

PR #527, opened 2025-01-13
516 check drivers version in khiops

bruno-at-orange:

PR #533, opened 2025-01-15
Check drivers version in khiops (#527)

bruno-at-orange:
* Add Utils.h/cpp in Norm

Move GetMajorVersion(const ALString& sFullVersion) and GetMinorVersion(const ALString& sFullVersion) from Learning library to a new file Utils.h in Norm

* Check the versions of the drivers when loading them

* Drivers messages are warnings


marcboulle (approved):
LGTM

PR #537, opened 2025-01-17
Update khiops_env: remove -hostfile and -n flags

bruno-at-orange:
Changes made to improve the usability on K8s:

- remove the KHIOPS_HOST_FILE: the host flag is not mandatory, it can be replaced by mpi environment variables
- by default, the -n flag is removed for openmpi, it will be easier to use on K8s

Refactoring:

we use _MPIEXEC to set the path of mpiexec (different on conda and system wide) And after all tests, we set KHIOPS_MPI_COMMAND. In tha way, on conda or not we're sure to have the same command line

### DIFF
--- a/packaging/install.cmake
+++ b/packaging/install.cmake
@@ -91,9 +91,9 @@ if(UNIX)
 
   endif(IS_CONDA)
 
-  # replace MPIEXEC MPIEXEC_NUMPROC_FLAG and MPI_IMPL KHIOPS_MPI_EXTRA_FLAG ADDITIONAL_ENV_VAR
+  # replace MPIEXEC MPIEXEC_NUMPROC_FLAG and MPI_IMPL MPI_EXTRA_FLAG ADDITIONAL_ENV_VAR
   if("${MPI_IMPL}" STREQUAL "openmpi")
-    set(KHIOPS_MPI_EXTRA_FLAG "--allow-run-as-root")
+    set(MPI_EXTRA_FLAG "--allow-run-as-root")
     set(KHIOPS_MPI_QUIET "--quiet")
     set(ADDITIONAL_ENV_VAR "export OMPI_MCA_btl_vader_single_copy_mechanism=none # issue on docker")
     set(ADDITIONAL_ENV_VAR_DISPLAY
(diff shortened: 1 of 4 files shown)


packaging/install.cmake
```diff
@@ -126,14 +123,26 @@ if(UNIX)
     set(MODL_NAME "MODL")
   endif()
 
+  # For all mpi implementation except openmpi, we compute the proc number (with openmpi, the -n flag is not mandatory)
+  if(NOT "${MPI_IMPL}" STREQUAL "openmpi")
```
popescu-v, line 127: I would have more explicitly written `if ("${MPI_IMPL}" STREQUAL "mpich")`, because in other places of this file we follow this approach of switching on the MPI implementation with a positive condition. And in these cases we only have "openmpi" and "mpich" as checked implementations anyway.

packaging/install.cmake
```diff
@@ -102,13 +102,10 @@ if(UNIX)
       set(ADDITIONAL_ENV_VAR "${ADDITIONAL_ENV_VAR}\nexport PSM3_DEVICES=self # issue on rocky linux")
       set(ADDITIONAL_ENV_VAR_DISPLAY "${ADDITIONAL_ENV_VAR_DISPLAY}\n    echo PSM3_DEVICES \"$PSM3_DEVICES\"")
     endif()
-    set(MPIEXEC_HOSTFILE_FLAG "--hostfile")
-    file(READ ${PROJECT_SOURCE_DIR}/packaging/linux/common/khiops_env/export_env_variables.sh EXPORT_ENV_VARIABLES)
   elseif("${MPI_IMPL}" STREQUAL "mpich")
```
popescu-v, line 105: IMHO there should be an `else` here to match any other implementation but OpenMPI (including Intel MPI which is just (like) MPICH IIUC).

And thus we would match the "OpenMPI vs anything else" approach that is used below in the current PR, e.g. on line 127.

packaging/install.cmake
```diff
@@ -126,14 +123,26 @@ if(UNIX)
     set(MODL_NAME "MODL")
   endif()
 
+  # For all mpi implementation except openmpi, we compute the proc number (with openmpi, the -n flag is not mandatory)
+  if(NOT "${MPI_IMPL}" STREQUAL "openmpi")
+    # Replace the path of _getprocnumber in set_proc_number.in (with the variable GET_PROC_NUMBER_PATH)
```
popescu-v: s/_getprocnumber/_khiopsgetprocnumber/

packaging/install.cmake
bruno-at-orange: I disagree, the fix for apple line 108 is specific to mpich

packaging/install.cmake
bruno-at-orange: done

packaging/install.cmake
popescu-v: OK, got it.

packaging/install.cmake
popescu-v, line 127: How would distributed computing work for MPICH (also natively supported), where we are obliged to set the `-n`?

popescu-v (approved):
LGTM

packaging/install.cmake
bruno-at-orange: from [Using_the_Hydra_Process_Manager](https://github.com/pmodels/mpich/blob/main/doc/wiki/how_to/Using_the_Hydra_Process_Manager.md)
```bash
mpiexec -f hosts -n 4 ./app
```
On a single node, the "-n" flag is mandatory. When the host file is set, I am not sure.

PR #540, opened 2025-01-21
Update khiops_env: remove -hostfile and -n flags (#537)

bruno-at-orange:
backport from #537 

Changes made to improve the usability on K8s:

- remove the KHIOPS_HOST_FILE: the host flag is not mandatory, it can be replaced by mpi environment variables
- by default, the -n flag is removed for openmpi, it will be easier to use on K8s

Refactoring: we use _MPIEXEC to set the path of mpiexec (different on conda and system wide) And after all tests, we set KHIOPS_MPI_COMMAND. In tha way, on conda or not we're sure to have the same command line

### DIFF
--- a/packaging/install.cmake
+++ b/packaging/install.cmake
@@ -91,9 +91,9 @@ if(UNIX)
 
   endif(IS_CONDA)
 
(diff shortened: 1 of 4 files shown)


popescu-v (approved):
LGTM

PR #549, opened 2025-01-27
Change conda package format to tar.bz2

bruno-at-orange:
Since the new package format is .conda we switch back to tar.bz2 to make minor changes in the CI

PR #551, opened 2025-01-28
Change conda package format to tar.bz2 (#549)

bruno-at-orange:
Since the new package format is .conda we switch back to tar.bz2 to make minor changes in the CI


popescu-v (approved):
LGTM

PR #571, opened 2025-02-05
518 remove compiler warnings on linux with gcc

### LINKED ISSUE #518: Remove compiler warnings on Linux with gcc

bruno-at-orange:
*  Disable spurious warning stringop-overflow

There is a false positive stringop-overflow warning on gcc 11
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106297
We add pragmas to disable this warning on minimum code lines

*  Remove gcc 11 Release warning: ‘this’ pointer is null [-Wnonnull]
*  The warning "reference to temporary [-Wreturn-local-addr]" is not handle in this PR because it is handle in another branch


marcboulle (approved):
LGTM

PR #649, opened 2025-03-26
Remove obsolete ubuntu 20.04 from actions

bruno-at-orange:


.github/workflows/conda.yml
```diff
@@ -92,7 +92,6 @@ jobs:
       fail-fast: false
       matrix:
         env:
-          - {os: ubuntu-20.04, os-family: linux-64}
```
popescu-v, line 95: IMHO, we need to keep testing the Conda packages on Ubuntu 20.04. To this end, I think we need to:

- use `runs-on: ubuntu-latest` everywhere applicable
- use `container: ${{ matrix.setup.os }}` everywhere applicable; thusly, we can use (brand new, almost empty) Docker containers for these tests (for example, `khiopstest-ubuntu20.04:latest` instead of the existing `khiopsdev-...`).

Thus we would run _the containers_ on the `ubuntu-latest` GithHub runners.

popescu-v (changes requested):
IMHO, we need to (see also the comment):
- systematically use `runs-on: ubuntu-latest` for all Linux runs, with `container: ...` according to the desired OS (the same as it is done in the `pack-debian.yml` workflow);
- keep testing the Conda packages on Ubuntu 20.04 (as before).

However, I also think we need to check whether, for tests, we can use GitHub-provided Docker images (as it is currently done in the `pack-{rpm,debian}.yml` workflows), or we should provide (and maintain) our own, thus make sure that:
- we would mitigate the risk of having GitHub stopping support for them;
- we would have total control over what is inside these images (that is, not much should be there, in order to match target test environments).
But this second part could be done in a different PR, perhaps.

bruno-at-orange:
superseded by PR #653

PR #653, opened 2025-03-27
Fix conda packaging

bruno-at-orange:
- Updating the conda package to account files with changing locations (since 85d2e2e055b3a5facc956bda57146fe0754b1d90)
- Use dockers to run tests (remove ubuntu 20.04 runners)
- fix new bug in windows/conda: we have to initialize Visual Studio environment before building


popescu-v (approved):
LGTM

PR #658, opened 2025-04-01
Use local build Conda channel with the highest priority

popescu-v:
Make sure that the locally-built khiops-core package is installed, rather than a possible khiops-core package situated on conda-forge.


bruno-at-orange (approved):
LGTM

PR #660, opened 2025-04-01
Unset Visual Studio Environment Variables on conda

bruno-at-orange:
An earlier fix e06af9d83ef was required because the Windows compiler was not found. This fix is no longer necessary and also causes a bug. This may be due to an instability in the Windows Runner. This commit is a revert from  e06af9d83ef


popescu-v (approved):
LGTM

PR #663, opened 2025-04-01
Fix conda tests in ci

popescu-v:
Various backports from v11 (cherry-picked commits from various V11 branches):

- use Docker images for testing the Conda package on Ubuntu 20.04;
- test the Conda package on Rocky 8 and 9;
- fix Python version to 3.12 when setting-up Miniconda for tests;
- download Conda build artifact before setting-up Miniconda; on Python 3.12, Miniconda reads the high-priority channel contents upon set-up.


bruno-at-orange (approved):
LGTM

PR #664, opened 2025-04-02
Fix sort algorithm: singleton and differents separators

### LINKED ISSUE #650: Border case? : sorting a datatable with duplicate lines produces an invalid output file
#### Context
Prior to a multi-tables TrainPredictor (Root datatable + Additional datatable), we need to sort the tables.

In a specific customer case, the additional table contains duplicates (the whole line can be found many times).

In this case, the sort does not behave correctly :
- the process ends without any error
- but in the output file we have a mix of lines with tab separated fields (expected) and comma separated fields (this is the input format and is unexpected).

```
(base) thierry@thierry-precision5570:~/Khiops/HECTOR$ grep -v "," flash_360_pseudonymization.csv_sorted | head -3
id_client	box_type	device_mac	event_family	event_type	event_subtype	nb_event	duration
000031433	Livebox 6	5E:0F:39:5A:89:F5	Wifi	RSSI Faible < -75dB	Wifi 5GHz	1	0.0
000031433	Livebox 6	26:07:6E:FD:49:25	Wifi	SSW Handover	Wifi 2GHz	1	0.0
```
```
(base) thierry@thierry-precision5570:~/Khiops/HECTOR$ grep "," flash_360_pseudonymization.csv_sorted | head -3
200231285,Livebox 6,4F:1F:8D:92:49:A0,Wifi,Connexion equipement,Wifi 2GHz,1,0.0
200231285,Livebox 6,4F:1F:8D:92:49:A0,Wifi,Connexion equipement,Wifi 2GHz,1,0.0
200231285,Livebox 6,4F:1F:8D:92:49:A0,Wifi,Connexion equipement,Wifi 2GHz,1,0.0
```

The output file is thus not valid and will produce error later if given to khiops as an input file.

#### Questions/Ideas
Many hypothesis were tested to find the root cause but all were dead-ends so far
	- empty fields : (cf ,,) 
	- duration under the scientific notation ? : 5.7870370370370366E-5

The only way to avoid totally the problem is to remove duplication in this additional datatable.
```
sort flash_360_pseudonymization.csv|uniq > flash_360_pseudonymization_uniq.csv

```

**Should we ask the customer to preprocess his additional datatable to remove the duplicates ?**

### Additional information
- Khiops version 1.3.0

- the khiops dictionary (dictionnaire.kdic) - [dictionnaire.kdic.txt](https://github.com/user-attachments/files/19482198/dictionnaire.kdic.txt)
- the pre-processed additional datatable (flash_360_pseudonymization.csv) compressed with BZ2 : it was "pseudonymized-with-format-preserved" to hide the customer MSISDN and MAC address - too big unfortunately to be attached to this issue and the binary format is not supported.
- the khiops scenario (khiops_sort.txt) - [khiops_sort.txt](https://github.com/user-attachments/files/19482167/khiops_sort.txt)

)

bruno-at-orange:
The singleton chunks don't need to be sorted, but if the input and output file separators are different, they need to be rewritten to change the file separator.

The rewrite is handled in the slaveProcess: it is the same process used for the chunk to be sorted, the only difference being that the keys are not sorted if it is a singleton. This can be optimised, of course, but singletons are rare, and optimising would make the code (which is already big enough) more complex.

### DIFF
--- a/src/Learning/KWDataUtils/KWChunkSorterTask.cpp
+++ b/src/Learning/KWDataUtils/KWChunkSorterTask.cpp
@@ -19,6 +19,7 @@ KWChunkSorterTask::KWChunkSorterTask()
 	// Variables en entree des taches
 	DeclareTaskInput(&input_nBucketIndex);
 	DeclareTaskInput(&input_svFileNames);
+	DeclareTaskInput(&input_bSingleton);
 
 	// Variables en sortie des taches
 	DeclareTaskOutput(&output_nBucketIndex);
@@ -32,6 +33,7 @@ KWChunkSorterTask::KWChunkSorterTask()
 	DeclareSharedParameter(&shared_lBucketSize);
 	DeclareSharedParameter(&shared_cInputSeparator);
 	DeclareSharedParameter(&shared_cOutputSeparator);
+	DeclareSharedParameter(&shared_bSameSeparator);
 
 	shared_cInputSeparator.SetValue('\t');
 	shared_cOutputSeparator.SetValue('\t');
@@ -135,7 +137,7 @@ boolean KWChunkSorterTask::Sort()
 	require(lBucketSize > 0);
 	bOk = Run();
 
-	// Nettoyage des fihiers de sortie en cas d'erreur
+	// Nettoyage des fichiers de sortie en cas d'erreur
 	if (not bOk)
 	{
 		for (i = 0; i < GetBuckets()->GetBucketNumber(); i++)
@@ -322,6 +324,7 @@ boolean KWChunkSorterTask::MasterInitialize()
 	shared_bHeaderLineUsed = bIsInputHeaderLineUsed;
 	shared_bOnlyOneBucket = buckets->GetBucketNumber() == 1;
 	shared_lBucketSize = lBucketSize;
+	shared_bSameSeparator = shared_cInputSeparator.GetValue() == shared_cOutputSeparator.GetValue();
 	return true;
 }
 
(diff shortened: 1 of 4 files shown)


src/Learning/KWDataUtils/KWChunkSorterTask.cpp
```diff
@@ -646,8 +646,12 @@ boolean KWChunkSorterTask::SlaveProcess()
 		// Tri des lignes extraites
 		if (bOk and not TaskProgression::IsInterruptionRequested())
 		{
-			oaKeyLines.SetCompareFunction(KWKeyLinePairCompare);
-			oaKeyLines.Sort();
+			// Si c'est un sigleton, on ne trie pas, on veut juste changer le separateur du fichier
```
marcboulle, line 649: sigleton -> singleton

src/Learning/KWDataUtils/KWChunkSorterTask.cpp
```diff
@@ -646,8 +646,12 @@ boolean KWChunkSorterTask::SlaveProcess()
...
+			if (not(input_bSingleton and shared_bSameSeparator))
```
marcboulle, line 650: if (not input_bSingleton)

marcboulle (approved):
LGTM
Petite correction

src/Learning/KWDataUtils/KWChunkSorterTask.cpp
bruno-at-orange: done

src/Learning/KWDataUtils/KWChunkSorterTask.cpp
bruno-at-orange: done

PR #666, opened 2025-04-02
Fix sort algorithm: singleton and differents file separators

bruno-at-orange:
The singleton chunks don't need to be sorted, but if the input and output file separators are different, they need to be rewritten to change the file separator.

The rewrite is handled in the slaveProcess: it is the same process used for the chunk to be sorted, the only difference being that the keys are not sorted if it is a singleton. This can be optimised, of course, but singletons are rare, and optimising would make the code (which is already big enough) more complex.


marcboulle (approved):
LGTM

ATTENTION: ne pas merger toute de suite, la V10.7.0-b.0 n'est pas encore valide

PR #670, opened 2025-04-03
Fix Unit Test framework

bruno-at-orange:
MPI Unit Tests were actually not launched with mpi. A (forgotten) cherry-pick from dev fixes it

### DIFF
--- a/.github/workflows/pack-debian.yml
+++ b/.github/workflows/pack-debian.yml
@@ -13,8 +13,13 @@ on:
   pull_request:
     paths:
       - '**CMakeLists.txt'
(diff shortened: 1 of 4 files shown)


popescu-v (commented):
Technically, LGTM for the scope of the current PR.
However, before approving, I wonder whether we should test in the CI that MPI tests are launched with `mpiexec` and fail the pipeline otherwise (thusly capturing potential regressions on this).

popescu-v (approved):
LGTM, can be merged after https://github.com/KhiopsML/khiops/pull/766.

PR #673, opened 2025-04-03
Pin the windows assets release to 10.7.0

### LINKED ISSUE #662: Pin assets version
We need to pin the version of the windows assets in dev-v11 and dev-v10
https://github.com/KhiopsML/khiops/blob/399605e10231e3f9468466a21e72ae0ce020410e/.github/workflows/pack-nsis.yml#L44

bruno-at-orange:
close #662

### DIFF
--- a/.github/workflows/pack-nsis.yml
+++ b/.github/workflows/pack-nsis.yml
@@ -45,7 +45,7 @@ jobs:
         uses: robinraju/release-downloader@v1.9
         with:
           repository: khiopsml/khiops-win-install-assets
-          latest: true
+          tag: 10.7.0
       - name: Extract Windows installer assets and load assets-info.env
         shell: bash
         run: |


bruno-at-orange:
The release process is updated with:
- [ ] Update the windows assets tag/version on the windows installer action (pack-nsis.yml)

.github/workflows/pack-nsis.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         uses: robinraju/release-downloader@v1.9
         with:
           repository: khiopsml/khiops-win-install-assets
-          latest: true
+          tag: 10.7.0
```
popescu-v, line 48: Could we use `11.0.0` as for `https://github.com/KhiopsML/khiops-samples` ? IMHO, we could:
- as the `10.7.x` tags are used for beta releases rather than official releses, right?
- if there is no pending upgrade to the assets, except for bug fixes, which could be handled by bumping the patch field of the version tag, right?

Or, otherwise, use a beta-like tag, e.g. `10.7.0-b.0`, as for `khiops-core`, to emphasize that this is just a beta release?

popescu-v:
Superseded by https://github.com/KhiopsML/khiops/pull/686.

Issue #646, opened 2025-03-21
Handle Werror=strict-prototypes with KNI


bruno-at-orange:
#### Description
A project using KNI uses the compilation flag "-Werror=strict-prototypes". Then gcc fails on KNI because functions without parameters should set "void" as parameter. See [stackoverflow](https://stackoverflow.com/questions/693788/is-it-better-to-use-c-void-arguments-void-foovoid-or-not-void-foo) for explanations.

lucaurelien:
à backporter sur la v10

bruno-at-orange:
Handled with #675 and #674

Issue #641, opened 2025-03-17
Update OS checking


bruno-at-orange:
When cross-compiling with [prplOS](https://gitlab.com/prpl-foundation/prplos/prplos), the build fails because  `__gnu_linux__` is not defined. We need to add `__linux__` in Portability.h. It is defined in prplos:
https://github.com/KhiopsML/khiops/blob/b6cc4c8fc2ef6c3f979132261a13332ada7bf29b/src/Norm/base/Portability.h#L40

marcboulle:
A back-porter

bruno-at-orange:
Handled with #675 and #674

PR #674, opened 2025-04-03
646 handle werrorstrict prototypes with kni

bruno-at-orange:
2 commits for 2 issues:
- Add "void" parameter functions without parameter in KNI:  #646 
- Add __linux__ as allowed os for Norm library:  #641

Notes:
- To be squashed under "Improve compilation on prplOS".
- To be ported on branch dev


marcboulle (approved):
LGTM
(et bientôt de même sur la branche dev)

PR #675, opened 2025-04-04
Improve compilation on prplOS (#674)

bruno-at-orange:
* Add "void" parameter functions without parameter in KNI

* Add __linux__ as allowed os for Norm library


marcboulle (approved):
LGTM

PR #713, opened 2025-06-16
fix sort algorithm: singleton vs sorted buckets (#700)

bruno-at-orange:
* fix sort algorithm: singleton vs sorted buckets

bucket->GetSorted() return false if bucket is a singleton (new behavior in 10.3.1) Buckets must be concatenated if they are sorted OR singletons.

* Remove files in KWChunkSorterTask in case of errors (previously done outside the task)

* Fix GetSortedLineNumber() in KWChunkSorterTask

Line number is computed during the sort in the SlaveProcess. Singleton chunks, are not processed by SlaveProcess unless the separators are different. Therefore, lines are not counted. We now increment the lines of the singleton in the MasterPrepareInput.

Lines number is now aggregated in KWSortedChunkBuilderTask: the built buckets contains the size of the bucket and the lines number. Minor refactoring in KWSortedChunkBuilderTask: we no longer use an ObjectDictionary to store and access buckets, instead we access the buckets directly from the shared object array.

### DIFF
--- a/src/Learning/KWDataUtils/KWChunkSorterTask.cpp
+++ b/src/Learning/KWDataUtils/KWChunkSorterTask.cpp
@@ -332,6 +332,9 @@ boolean KWChunkSorterTask::MasterPrepareTaskInput(double& dTaskPercent, boolean&
 {
 	KWSortBucket* bucketToSort;
 	int nTreatedBucketNumber;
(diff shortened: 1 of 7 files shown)


marcboulle (approved):
LGTM

PR #720, opened 2025-06-24
Signature of the windows installer

bruno-at-orange:
We use a github action from digicert to install smctl utility. We sign all binaries (as well as the dll and the installer). We use a test signature with a limited lifespan and a production signature. The production signature is only used for releases based on tags that are neither release candidate nor alphas

### DIFF
--- a/.github/workflows/pack-nsis.yml
+++ b/.github/workflows/pack-nsis.yml
@@ -15,6 +15,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref
     }}
   cancel-in-progress: true
+env:
+  KEYPAIR_TEST: KP_Khiops_Test
+  KEYPAIR_PROD: KP_Khiops_HSM
 jobs:
   build:
     outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/pack-nsis.yml
```diff
@@ -59,6 +62,44 @@ jobs:
...
+      - name: Set variables for signature
+        id: variables-used-by-smctl
+        run: |
+          echo "SM_HOST=${{ secrets.SM_HOST }}" >> "$GITHUB_ENV" 
+          echo "SM_API_KEY=${{ secrets.SM_API_KEY }}" >> "$GITHUB_ENV" 
+          echo "SM_CLIENT_CERT_FILE=D:\\Certificate_pkcs12.p12" >> "$GITHUB_ENV" 
+          echo "SM_CLIENT_CERT_PASSWORD=${{ secrets.SM_CLIENT_CERT_PASSWORD }}" >> "$GITHUB_ENV"
+
+          # The production key is only used for releases from tags that are not release candidates or alpha
+          if [[ "${{ github.ref_type }}" == "tag" ]] && [[ "${{ github.ref_name  }}" != *"-rc."* ]] && [[ "${{ github.ref_name  }}" != *"-a."* ]]; then
```
popescu-v, line 81: Is it intended that one uses the production certificate for beta pre-releases?
If so, then this seems OK.
If not, then the relevant condition on the `"-b."*` should be added.

However, the "release candidate" pre-release type is closer to an official release than a "beta" pre-release type AFAIU. Hence, using a more perennial certificate for beta pre-releases than for release candidates seems counterintuitive to me and needs to be justified IMHO.

.github/workflows/pack-nsis.yml
```diff
@@ -59,6 +62,44 @@ jobs:
...
+          else
+            echo "KEYPAIR=$KEYPAIR_TEST" >> "$GITHUB_ENV"
+            echo "::notice::Using Test key for signature"
+          fi
+        shell: bash
+      - name: Signing using keypair alias
+        run: |
+          OUTPUT=$(smctl sign --keypair-alias $KEYPAIR --config-file C/Users/RUNNER~1/AppData/Local/Temp/smtools-windows-x64/pkcs11properties.cfg --input build/windows-msvc-release/bin/MODL.exe)
+          echo $OUTPUT
+          echo $OUTPUT | grep -q SUCCESSFUL
```
popescu-v, line 93: This debugging lines are here to stay? I would perhaps consider removing them once the workflow is stabilized.

.github/workflows/pack-nsis.yml
```diff
@@ -59,6 +62,44 @@ jobs:
...
+          OUTPUT=$(smctl sign --keypair-alias $KEYPAIR --config-file C/Users/RUNNER~1/AppData/Local/Temp/smtools-windows-x64/pkcs11properties.cfg --input build/windows-msvc-release/bin/MODL_Coclustering.exe)
+          echo $OUTPUT
+          echo $OUTPUT | grep -q SUCCESSFUL
```
popescu-v, line 96: This debugging lines are here to stay? I would perhaps consider removing them once the workflow is stabilized.

.github/workflows/pack-nsis.yml
```diff
@@ -59,6 +62,44 @@ jobs:
...
+          OUTPUT=$(smctl sign --keypair-alias $KEYPAIR --config-file C/Users/RUNNER~1/AppData/Local/Temp/smtools-windows-x64/pkcs11properties.cfg --input build/windows-msvc-release/bin/_khiopsgetprocnumber.exe)
+          echo $OUTPUT
+          echo $OUTPUT | grep -q SUCCESSFUL
```
popescu-v, line 99: This debugging lines are here to stay? I would perhaps consider removing them once the workflow is stabilized.

.github/workflows/pack-nsis.yml
```diff
@@ -75,6 +116,12 @@ jobs:
             /DKHIOPS_SAMPLES_DIR=..\..\..\assets\samples `
             /DKHIOPS_DOC_DIR=..\..\..\assets\doc `
             khiops.nsi
+      - name: Signing installer
+        run: |
+          OUTPUT=$(smctl sign --keypair-alias $KEYPAIR --config-file C/Users/RUNNER~1/AppData/Local/Temp/smtools-windows-x64/pkcs11properties.cfg --input packaging/windows/nsis/khiops-${{ env.KHIOPS_VERSION }}-setup.exe)
+          echo $OUTPUT
+          echo $OUTPUT | grep -q SUCCESSFUL
```
popescu-v, line 124: This debugging lines are here to stay? I would perhaps consider removing them once the workflow is stabilized.

.github/workflows/pack-nsis.yml
```diff
@@ -75,6 +116,12 @@ jobs:
             /DKHIOPS_SAMPLES_DIR=..\..\..\assets\samples `
             /DKHIOPS_DOC_DIR=..\..\..\assets\doc `
             khiops.nsi
+      - name: Signing installer
```
popescu-v, line 120: Perhaps I'm missing something, but shouldn't the uninstaller also be signed? If it is, where is this being done?

popescu-v (changes requested):
See the comments.

.github/workflows/pack-nsis.yml
bruno-at-orange: I prefer to see the smctl output in any case

.github/workflows/pack-nsis.yml
bruno-at-orange: I prefer to see the smctl output in any case

.github/workflows/pack-nsis.yml
bruno-at-orange: I prefer to see the smctl output in any case

.github/workflows/pack-nsis.yml
bruno-at-orange: I prefer to see the smctl output in any case

.github/workflows/pack-nsis.yml
bruno-at-orange: Only on the dev branch, it hasn'y been backported

.github/workflows/pack-nsis.yml
popescu-v: Ok.

.github/workflows/pack-nsis.yml
popescu-v: Ok.

.github/workflows/pack-nsis.yml
bruno-at-orange: comment added

popescu-v (approved):
LGTM, modulo the calls into the signing commands themselves.

PR #728, opened 2025-06-27
Signing on GitHub backport

bruno-at-orange:
- Backport of the signature process done on dev-v10
- Modify nsi file to handle signature of the uninstaller inside the NSIS build.
- Unfortunately this last signature fails with the message "There were no files found for signing" but this problem does not affect the installation. I suggest to merge these features anyway and to create an issue about it.


marcboulle (approved):
LGTM

PR #751, opened 2025-07-21
Add branches in kni tutorial

bruno-at-orange:


marcboulle (approved):
Just lower case one char:
"Application Examples are available" -> "Application examples are available"

PR #757, opened 2025-07-23
723 do not install pdf docs in windows installer

### LINKED ISSUE #723: Do not install pdf docs in windows installer
Repos asset à mettre en phase

bruno-at-orange:
The documentation is no longer packaged in the Windows installer.
Furthermore, the [assets repository](https://github.com/KhiopsML/khiops-win-install-assets) has been cleaned up (starting from tag 11.0.1). We removed:
- documentation
- Visualization tools
- samples

Only `msmpi` and `jre` remain. The samples and visualization tools are downloaded directly from their repositories before packaging.

### DIFF
--- a/.github/workflows/pack-nsis.yml
+++ b/.github/workflows/pack-nsis.yml
@@ -47,13 +47,27 @@ jobs:
         uses: robinraju/release-downloader@v1.9
         with:
           repository: khiopsml/khiops-win-install-assets
-          tag: 11.0.0
+          tag: 11.0.1
       - name: Extract Windows installer assets and load assets-info.env
         shell: bash
         run: |
           assets_tar_gz=$(ls khiops-win-install-assets*.tar.gz)
           tar -zxvf "$assets_tar_gz"
           cat assets/assets-info.env >> "$GITHUB_ENV"
+      # We download the latest version of visualization tools: 
+      # - They are backward compatible.
+      # - The process is made easier by getting automatically the latest version
+      - name: Download latest releases of khiops-*visualization
+        shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh --repo https://github.com/KhiopsML/kc-electron release download --pattern *.exe -D visualizations 
+          gh --repo https://github.com/KhiopsML/kv-electron release download --pattern *.exe -D visualizations
+          cd visualizations
+          ls *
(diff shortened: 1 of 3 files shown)


.github/workflows/pack-nsis.yml
```diff
@@ -47,13 +47,24 @@ jobs:
...
           repository: khiopsml/khiops-win-install-assets
-          tag: 11.0.0
+          tag: 11.0.1
       - name: Extract Windows installer assets and load assets-info.env
         shell: bash
         run: |
           assets_tar_gz=$(ls khiops-win-install-assets*.tar.gz)
           tar -zxvf "$assets_tar_gz"
           cat assets/assets-info.env >> "$GITHUB_ENV"
+      - name: Download latest releases of khiops-*visualization
```
popescu-v, line 62: IMHO, keeping the visualization and covisualization tool versions explicit (via environment variables, with default values possibly overridable by workflow inputs) would ensure the same traceability as before. Using "latest" would make us lose this traceability IMHO.

popescu-v (changes requested):
I would keep the visualization and covisualization tool versions explicit (as it was the case, indirectly, before) - see the comment.

.github/workflows/pack-nsis.yml
bruno-at-orange: I am not agree, when we release  Khiops we have to install the last versions of the visualization tools. It is extra work to pin the latest version each time. The traceability of the visualization tools version doesn't matter especially since the visualization tools have auto-update mechanism. To discuss in meeting

marcboulle:
Ne pas oublier de prendre le tag 11.0.0 des samples

bruno-at-orange:
> Ne pas oublier de prendre le tag 11.0.0 des samples

C'est bon Cf. pack-nsis.yml:20

.github/workflows/pack-nsis.yml
popescu-v: OK, understood.
What I had in mind was especially the v10 / v11 dichotomy. But if this dichotomy is irrelevant to the visualization tools, this change seems OK to me.
However, as this change entails a difference wrt. the previous mechanism, I would add a comment along these lines:
- one in the commit message to emphasize that this behaviour change is inconsequential in practice;
- one in the workflow to emphasize that pinning the visualization tools doesn't matter.

bruno-at-orange: Comments added to commit note and yml

popescu-v: OK, good.

popescu-v (approved):
LGTM

PR #758, opened 2025-07-23
Update KHIOPS_API_MODE Environment Variable Help in khiops_env

### LINKED ISSUE #558: Update `KHIOPS_API_MODE` Environment Variable Help in `khiops_env`
#### Description
Currently, `KHIOPS_API_MODE` is documented thus:
```
KHIOPS_API_MODE: standard or api mode for the management of output result files created by Khiops
  In standard mode, the result files are stored in the train database directory,
  unless an absolute path is specified, and the file extension is forced if necessary.
  In api mode, the result files are stored in the current working directory, using the specified results as is.
  . default behavior if not set: standard mode
  . set to 'false' to force standard mode
  . set to 'true' to force api mode
```

#### Questions/Ideas
- A more accurate description of the effects of this variable would be:
```
KHIOPS_API_MODE: standard or api mode for the management of output result files created by Khiops
  In standard mode, the result files are stored relative to the train database directory and the report file
  extension is automatically added if not specified.
  In api mode, the result files are stored relative to the current working directory and the report file name
  is used as specified.
  In both modes, if an absolute path to the report file is specified, then the result files are stored in the
  same directory as the report file.
  . default behavior if not set: standard mode
  . set to 'false' to force standard mode
  . set to 'true' to force api mode
```
  - The key semantic changes are that:
    - in all cases, if absolute path to the report file is given, then all the result files are stored in the directory of the report file. 
    - in all cases, result files are stored _relative_ to a directory which depends on the mode, not necessarily directly _in_ that directory; this accommodates relative paths to the report file.

### Context
- Khiops version: >= 11

bruno-at-orange:
close #558


popescu-v (approved):
LGTM

PR #770, opened 2025-08-04
734 signature fails for windows uninstaller

### LINKED ISSUE #734: Signature fails for windows uninstaller
The signature embedded in the .nsi file fails with the message "There were no files found for signing".

bruno-at-orange:
NSIS builds the uninstaller with the '.tmp' extension. The tool provided by DigiCert, smctl, fails to sign a 'tmp' file. The workaround is to use jsign to sign the uninstaller.

close #734

Improvements outside the scope of the issue:
- Add a parameter to the workflow to enable testing of the signature process, even if the current branch is not a tag (I will use it when PR will be approved).
- Improve smctl usage: multiple files can be signed at once. The return code is 1 in case of errors.

### DIFF
--- a/.github/workflows/pack-nsis.yml
+++ b/.github/workflows/pack-nsis.yml
@@ -2,6 +2,12 @@
 name: Windows Installer
 on:
   workflow_dispatch:
+    inputs:
+      signature-activation:
+        type: boolean
+        required: true
+        default: false
+        description: Sign installer and binaries
   pull_request:
     paths:
       - '**CMakeLists.txt'
@@ -19,8 +25,14 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref
     }}
   cancel-in-progress: true
+# Environment variables:
+# - SAMPLES_VERSION: tag of the sample repository packed in the installer
+# - ASSETS_VERSION: tag of the khiops-win-install-assets repository packed in the installer
+# - KEYPAIR is used by smctl 
+# - CERTIFICATE is used by jsign
 env:
   KEYPAIR: KP_Khiops_HSM
+  CERTIFICATE: Cert_Khiops
   SAMPLES_VERSION: 11.0.0
   ASSETS_VERSION: 11.0.1
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/pack-nsis.yml
```diff
@@ -19,8 +24,14 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref
     }}
   cancel-in-progress: true
+# Environment variables:
+# - SAMPLES_VERSION: tag of the sample repository packed in the installer
+# - ASSETS_VERSION: tah of the windows-asset repository packed in the installer
```
popescu-v: s/tah/tag/
s/windows-asset/khiops-win-install-assets/

.github/workflows/pack-nsis.yml
```diff
@@ -98,22 +109,22 @@ jobs:
           echo "SM_CLIENT_CERT_PASSWORD=${{ secrets.SM_CLIENT_CERT_PASSWORD }}" >> "$GITHUB_ENV"
 
           # The packages are signed only for releases from tags that are not release candidates or alpha
-          # Note: We sign  for the beta releases because they are distributed to external users 
-          if [[ "${{ github.ref_type }}" == "tag" ]] && [[ "${{ github.ref_name  }}" != *"-rc."* ]] && [[ "${{ github.ref_name  }}" != *"-a."* ]]; then
+          # Note: We sign for the beta releases because they are distributed to external users 
+          if [[ "${{ github.ref_type }}" == "tag" && "${{ github.ref_name  }}" != *"-rc."* && "${{ github.ref_name  }}" != *"-a."* ]] || ${{ inputs.signature-activation == 'true'}} ; then
```
popescu-v: Could we simplify this process and only sign on manual request? Do we really need to automate the signature for non-alpha and non-rc tags (i.e. we might have _internal_ betas at some point, which we wouldn't necessarily sign)?

.github/workflows/pack-nsis.yml
```diff
@@ -98,22 +109,22 @@ jobs:
...
             echo "SIGN=ON" >> "$GITHUB_ENV"
             echo "::notice::The signing process is ON"
           else
             echo "SIGN=OFF" >> "$GITHUB_ENV"
             echo "::notice::The signing process is OFF"
-          fi
+          fi
```
popescu-v: Drop trailing spaces (discard this change).

popescu-v (changes requested):
Just a small request for simplification (see comment).

.github/workflows/pack-nsis.yml
bruno-at-orange: done

.github/workflows/pack-nsis.yml
bruno-at-orange: done

.github/workflows/pack-nsis.yml
bruno-at-orange: OK

bruno-at-orange:
Code is up to date. I added an entry "Sign windows installer" in the [wiki](https://github.com/KhiopsML/khiops/wiki/Release-Process)

.github/workflows/pack-nsis.yml
```diff
@@ -96,43 +108,27 @@ jobs:
...
-        if: env.SIGN == 'ON'
+        if: ${{ inputs.signature-activation }}
+        run: |
+          echo "${{ secrets.SM_CLIENT_CERT_FILE_B64 }}" | base64 --decode > /d/Certificate_pkcs12.p12
+        shell: bash
+      - name: Signing binaries using keypair alias
+        if: ${{ inputs.signature-activation }}
         run: |
-          echo "${{ secrets.SM_CLIENT_CERT_FILE_B64 }}" | base64 --decode > /d/Certificate_pkcs12.p12 
+          smctl sign --exit-non-zero-on-fail --keypair-alias $KEYPAIR --config-file C/Users/RUNNER~1/AppData/Local/Temp/smtools-windows-x64/pkcs11properties.cfg --input build/windows-msvc-release/bin/
```
popescu-v: So, now we sign the whole contents of the built `/bin` directory instead of explicitly signing each of the relevant files?
If this is the intended behavior, then I would add a comment stating so, perhaps listing the relevant files that are being signed.

popescu-v (changes requested):
Small request for comment as per file selection for signature change (see the remaining open comment).

.github/workflows/pack-nsis.yml
bruno-at-orange: Yes it signs all files in the `bin` directory. But actually, KNITransfer.exe doesn't need  to be signed. So I come back to list each file.

popescu-v (approved):
LGTM, but IMHO the text "-Use directory as input" should be removed from commit https://github.com/KhiopsML/khiops/pull/770/commits/0f00f1da0974000d77d5f19323200e68ad7c5a09's commit message.

PR #809, opened 2025-11-07
743 add functionality to compute and store delta variables for numerical categorical and temporal data

### LINKED ISSUE #743: Add functionality to compute and store delta variables for numerical, categorical, and temporal data
We need to implement a feature that allows adding new variables to a table based on existing variables, specifically:

- Delta time: Calculate `T(n) - T(n-1)` for each row, representing the time difference between the current row and the previous one.
- Delta numerical: Calculate `N(n) - N(n-1)` for each row, representing the change in the numerical variable between consecutive rows.
- Change in categorical variable: Determine whether the categorical variable `C(n)` has changed compared to `C(n-1)` (e.g., a boolean flag or a new variable indicating change).

This functionality will enable better analysis of data trends and transitions over time or sequence.

marcboulle:


carinehue (approved):
LGTM

PR #822, opened 2025-11-21
815 mpi freezes in the snb when a user interruption occurs

### LINKED ISSUE #815: MPI freezes in the SNB when a 'user interruption' occurs.
Un comportement de freeze en cas d'interruption dans la partie "worker initialisation" du SNB

Ce problème provient potentiellement du framework parallèle, et peut être investigué dans la branche en cours?

Jeu de test:
- LearningTest\TestKhiops\MTClassification\Accidents
- uniquement le cas le plus long (1000 attributs construits)
- sans les arbres
- en release parallèle
- interruption utilisateur au moment de la partie "worker initialisation" du SNB
- freeze non systématique, mais assez rapide à reproduire (2 à 3 essais)

bruno-at-orange:
- Bug fix:
In case of user interruption, the master receives the end of the slave initialization.
If the slave initialization fails, the state of the slave is set to 'ERROR' and no other message is expected from this slave.
- Improvements:
  - TaskProgression is set to interrupted when PLMPISlaveProgressionManger detects an interruption.
This removes unnecessary calls to the timer and to PLMPISlaveProgressionManger.
  - Remove unused method PLMPISlave::WaitForMessage()
- Code cleanup:
The NotifyAbnormalExit method was removed because it was only used once. This method is no longer necessary because the master knows if initialization has failed. This simplifies the code and eliminates an MPI message when slaveInitialize fails.


marcboulle (approved):
LGTM

PR #856, opened 2025-12-10
Fix non-MPI CMake config on Fedora-like distros

tristandeb1:
Only check the presence of MPI_BIN and MPI_SUFFIX if MPI option is enabled.


bruno-at-orange (approved):
LGTM

PR #874, opened 2026-01-14
Use new digicert action

### LINKED ISSUE #735: Add a new action to sign binaries
Currently, each repository manage the signature with the same code lines. It would be better to create an action in order to simplify the possible future updates of the signature process.

bruno-at-orange:
We can not share an action between projects without sharing digicert secrets. Indeed, the new action have to access to the secrets. So any project using this new action can sign wit our certificate.

However, the action [digicert/ssm-code-signing](https://github.com/digicert/ssm-code-signing) is deprecated. We switch to the new one [digicert/code-signing-software-trust-action](https://github.com/digicert/code-signing-software-trust-action).

### DIFF
--- a/.github/workflows/pack-nsis.yml
+++ b/.github/workflows/pack-nsis.yml
@@ -103,8 +103,8 @@ jobs:
           override-flags: -DTESTING=OFF -DBUILD_JARS=ON
       - name: Setup SM_CLIENT_CERT_FILE from base64 secret data
         run: |
-          echo "SM_CLIENT_CERT_FILE=D:\\Certificate_pkcs12_test.p12" >> "$GITHUB_ENV"
-          echo "${{ secrets.SM_CLIENT_CERT_FILE_B64 }}" | base64 --decode > /d/Certificate_pkcs12_test.p12
+          echo "SM_CLIENT_CERT_FILE=D:\\Certificate_pkcs12.p12" >> "$GITHUB_ENV"
+          echo "${{ secrets.SM_CLIENT_CERT_FILE_B64 }}" | base64 --decode > /d/Certificate_pkcs12.p12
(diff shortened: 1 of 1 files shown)


.github/workflows/pack-nsis.yml
```diff
@@ -101,32 +101,22 @@ jobs:
...
           override-flags: -DTESTING=OFF -DBUILD_JARS=ON
-      - name: Set variables for signature
-        id: variables-used-by-smctl
+      - name: Setup SM_CLIENT_CERT_FILE from base64 secret data
         run: |
-          echo "SM_HOST=${{ secrets.SM_HOST }}" >> "$GITHUB_ENV" 
-          echo "SM_API_KEY=${{ secrets.SM_API_KEY }}" >> "$GITHUB_ENV" 
-          echo "SM_CLIENT_CERT_FILE=D:\\Certificate_pkcs12.p12" >> "$GITHUB_ENV" 
-          echo "SM_CLIENT_CERT_PASSWORD=${{ secrets.SM_CLIENT_CERT_PASSWORD }}" >> "$GITHUB_ENV"
+          echo "SM_CLIENT_CERT_FILE=D:\\Certificate_pkcs12_test.p12" >> "$GITHUB_ENV"
```
popescu-v: Why is there `_test` in the certificate file name?

popescu-v (approved):
LGTM, although I did not check the `digicert/code-signing-software-trust-action` action itself.

.github/workflows/pack-nsis.yml
bruno-at-orange: It was to verify that the certificate password is not linked to the certificate name. In other words, we can name the certificate whatever we want. I remove the test suffix now.

PR #905, opened 2026-02-02
Add a trace in kht_test to indicate the use of khiops_env

bruno-at-orange:
Add a trace to indicate if kit_test uses khiops_env. It looks like:
```bash
starting Test TestKhiops Standard Adult (processes: 18, platform: Linux, khiops_env: True)
  exe: /usr/bin/MODL_openmpi
```


marcboulle (approved):
LGTM

PR #932, opened 2026-02-18
Package pip

bruno-at-orange:
- Update khisto installation to be compatible with PIP
- CMake improvments

### DIFF
--- a/packaging/install.cmake
+++ b/packaging/install.cmake
@@ -252,8 +252,18 @@ else(UNIX)
 endif(UNIX)
 
 # ######################################## khisto installation
-install(TARGETS khisto RUNTIME DESTINATION bin COMPONENT KHISTO)
-install(
-  FILES ${PROJECT_SOURCE_DIR}/LICENSE
-  DESTINATION doc
-  COMPONENT KHISTO)
+if(IS_PIP)
+  install(TARGETS khisto RUNTIME DESTINATION ${SKBUILD_SCRIPTS_DIR} COMPONENT KHISTO)
+elseif(UNIX)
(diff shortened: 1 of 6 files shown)


packaging/install.cmake
```diff
@@ -252,8 +252,18 @@ else(UNIX)
 endif(UNIX)
 
 # ######################################## khisto installation
-install(TARGETS khisto RUNTIME DESTINATION bin COMPONENT KHISTO)
-install(
-  FILES ${PROJECT_SOURCE_DIR}/LICENSE
-  DESTINATION doc
-  COMPONENT KHISTO)
+if(IS_PIP)
+  install(TARGETS khisto RUNTIME DESTINATION ${SKBUILD_SCRIPTS_DIR} COMPONENT KHISTO)
```
popescu-v, line 250: Shouldn't we also put the license file in the Pip package?

popescu-v (changes requested):
See the comment.

Shouldn't we update the CI to build the Pip package (even though it's not yet pushed on PyPI)?

packaging/install.cmake
bruno-at-orange: no it is already done in the `pyproject.toml` I will comment this

bruno-at-orange:
the Pip package is built in khisto (and khiops-python in the future) not in the khiops repo

popescu-v (approved):
LGTM. The case when `IS_PIP` is true in `install.cmake` is / will be tested in `https://github.com/KhiopsML/khisto-python/`, namely after `https://github.com/KhiopsML/khisto-python/blob/1-package-khisto-python/.github/workflows/pack-pip.yml` is integrated into the `main` branch of `khisto-python`.

PR #957, opened 2026-03-11
726 remove conda generation

### LINKED ISSUE #726: Remove conda generation
Conda packages are now managed on conda-forge. We have to remove the conda build process from the khiops actions.

### LINKED ISSUE #903: Remove actions test-khiops-on-iris
The test-khiops-on-iris is always used in conjunction with run-standard-test. It is perhaps useless.

bruno-at-orange:
- Remove conda generation on te CI
- Remove the action test-khiops-on-iris close #903


.github/workflows/conda.yml
```diff
@@ -1,214 +0,0 @@
...
-        run: |
-          conda build --package-format=tar.bz2 --output-folder ./build/conda ./packaging/conda
-      - name: Upload conda packages artifact
-        uses: actions/upload-artifact@v7
-        with:
-          name: khiops-conda-${{ matrix.setup.os-family }}
-          path: ./build/conda
-          retention-days: 7
-  # Test Conda package on brand new environments
-  test:
```
popescu-v, line 89: I would have kept the `test` job to have the full target platform matrix. But the package download step needs to be changed so that the package on `conda-forge` is downloaded instead. See also https://github.com/KhiopsML/khiops-python/issues/503.

packaging/conda/conda_build_config.yaml
```diff
@@ -1,32 +1,9 @@
...
-# - requires conda-forge
-# Windows:
-# - msmpi=10.1.1 is the latest as of this commit
+# Setting the mpi implementation
+# msmpi is not in the global pinning
 mpi:
-  - mpich=4.0.3  # [linux]
-  - mpich=3.4.3  # [osx]
-  - msmpi=10.1.1  # [win]
-c_compiler:
```
popescu-v, line 16: In order for the _local_ build to succeed, we also need to add:
```yaml
c_stdlib:
  - sysroot                     # [linux]
  - macosx_deployment_target    # [osx]
```
which are defined in the feedstock, in https://github.com/conda-forge/khiops-binaries-feedstock/blob/fda2323642fd30f03beca4e35ac96eaa4e427ac7/.ci_support/linux_64_.yaml#L1 for Linux and, respectively, https://github.com/conda-forge/khiops-binaries-feedstock/blob/fda2323642fd30f03beca4e35ac96eaa4e427ac7/.ci_support/osx_arm64_.yaml#L5, for MacOS ARM (identical for MACOS x86 in https://github.com/conda-forge/khiops-binaries-feedstock/blob/fda2323642fd30f03beca4e35ac96eaa4e427ac7/.ci_support/osx_64_.yaml#L5).

Otherwise, the local build fails with
```bash
The following package could not be installed
└─ c_linux-64 =* * does not exist (perhaps a typo or a missing channel).
```
See issue https://github.com/conda/conda-build/issues/5745 and comment https://github.com/conda/conda-build/issues/5745#issuecomment-3141449270. However, apparently the local build succeeds in our case even if we don't specify the `c_stdlib_version` (which is, however, indicated in the feedstock, in the same files mentioned above).

popescu-v (changes requested):
See the two comments, especially the second one.
I also wonder whether we shouldn't test the local Conda build in the CI to make sure the recipe provided here works.

.github/workflows/conda.yml
bruno-at-orange: OK, tests are in a new file: test-conda.yml

packaging/conda/conda_build_config.yaml
bruno-at-orange: the paragraph c_stdlib is added

bruno-at-orange:
> I also wonder whether we shouldn't test the local Conda build in the CI to make sure the recipe provided here works.

I don't think so. We provided this recipe for testing purposes, so it's not really a problem if it stops working at any time. We will correct it only when necessary.

.github/workflows/test-conda.yml
```diff
@@ -0,0 +1,78 @@
...
+          channels: conda-forge
+          conda-remove-defaults: true
+      - name: Install the Khiops executables in the Conda environment
+        run: |
+          if [ -z "${{ github.event.inputs.khiops_version }}" ]; then
+            echo "No version specified, installing the latest version of khiops-core"
+            conda install -c khiops-core/label/${{ github.event.inputs.label }}  khiops-core
+          else
+            echo "Installing khiops-core version ${{ github.event.inputs.khiops_version }}"
+            conda install -c khiops-core/label/${{ github.event.inputs.label }}  khiops-core=${{ github.event.inputs.khiops_version }}
```
popescu-v: I think it's `conda-forge/label/${{ github.event.inputs.label }}`.

.github/workflows/test-conda.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          miniforge-version: latest
+          python-version: '3.12'
+          channels: conda-forge
+          conda-remove-defaults: true
+      - name: Install the Khiops executables in the Conda environment
+        run: |
+          if [ -z "${{ github.event.inputs.khiops_version }}" ]; then
+            echo "No version specified, installing the latest version of khiops-core"
+            conda install -c khiops-core/label/${{ github.event.inputs.label }}  khiops-core
```
popescu-v: I think it's `conda-forge/label/${{ github.event.inputs.label }}`.

packaging/conda/conda_build_config.yaml
```diff
@@ -1,34 +1,14 @@
...
-  - 12  # [linux]
-  - 16  # [osx]
+  - mpich  # [linux or osx]
+  - msmpi =10.1.1  # [win]
 # We need MacOS SDK 10.10 to be able to build on macOS Intel
 # Download: https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX10.10.sdk.tar.xz
 # Decompress then to /opt: tar -zxvf MacOSX10.10.sdk.tar.xz -C /opt
 CONDA_BUILD_SYSROOT:
   - /opt/MacOSX10.10.sdk  # [osx and not arm64]
+c_stdlib:
```
popescu-v, line 12: I wonder whether we shouldn't also take `c_stdlib_version` from https://github.com/conda-forge/khiops-binaries-feedstock/blob/fda2323642fd30f03beca4e35ac96eaa4e427ac7/.ci_support/linux_64_.yaml#L3 and https://github.com/conda-forge/khiops-binaries-feedstock/blob/fda2323642fd30f03beca4e35ac96eaa4e427ac7/.ci_support/osx_arm64_.yaml#L7 (mac ARM) and https://github.com/conda-forge/khiops-binaries-feedstock/blob/fda2323642fd30f03beca4e35ac96eaa4e427ac7/.ci_support/osx_64_.yaml#L7 (mac x86) in order to match the `conda-forge` build environment to the maximum.

packaging/conda/meta.yaml
```diff
@@ -1,53 +1,38 @@
...
-  detect_binary_files_with_prefix: false # [osx]
-  {% endif %}
 
 requirements:
   build:
+    - {{ mpi }}  # [mpi == 'openmpi' and build_platform != target_platform]
     - cmake
     - ninja
     - {{ compiler('cxx') }}
+    - {{ stdlib("c") }}  # [not (osx and arm64)]
```
popescu-v: The selector is not on the `conda-forge` feedstock: why exclude this requirement for the local mac ARM Conda builds?

packaging/conda/meta.yaml
```diff
@@ -1,53 +1,38 @@
...
-  # khiops-core package
+# khiops-core package
   - name: khiops-core
     version: {{ version }}
     requirements:
       build:
         - cmake
         - ninja
         - {{ compiler('cxx') }}
+        - {{ stdlib("c") }}  # [not (osx and arm64)]
```
popescu-v: The selector is not on the `conda-forge` feedstock: why exclude this requirement for the local mac ARM Conda builds?

popescu-v (changes requested):
See the pending comments.
Also update the `packaging/conda/bld.bat` script:
- drop all `KNITRansfer` related instructions:
  - from the CMake build targets
  - from the executables that are copied to `%LIBRARY_BIN%`
- copy the KNI header file to `%LIBRARY_INC%`

packaging/conda/meta.yaml
bruno-at-orange: because it won't compile without it

bruno-at-orange:
> See the pending comments. Also update the `packaging/conda/bld.bat` script:
> 
> * drop all `KNITRansfer` related instructions:
>   
>   * from the CMake build targets
>   * from the executables that are copied to `%LIBRARY_BIN%`
> * copy the KNI header file to `%LIBRARY_INC%`

done. I've completely forgotten about Windows...

packaging/conda/meta.yaml
popescu-v: Maybe it's because we need to set `c_stdlib_version` as in https://github.com/conda-forge/khiops-binaries-feedstock/blob/52ef7add6a3d382f5c28ca84b2c74d3e1f641c3a/.ci_support/osx_arm64_.yaml#L7?

IMHO, it's important to be as close as possible to the feedstock, unless in obvious cases (like the `source` statement).

packaging/conda/meta.yaml
popescu-v: Maybe it's because we need to set `c_stdlib_version` as in https://github.com/conda-forge/khiops-binaries-feedstock/blob/52ef7add6a3d382f5c28ca84b2c74d3e1f641c3a/.ci_support/osx_arm64_.yaml#L7?

IMHO, it's important to be as close as possible to the feedstock, unless in obvious cases (like the `source` statement).

packaging/conda/bld.bat
```diff
@@ -3,24 +3,20 @@ REM Echo all output
 
 REM Build the Khiops binaries
 cmake -B build\conda -S . -D BUILD_JARS=OFF -D TESTING=OFF -D CMAKE_BUILD_TYPE=Release -G Ninja
-cmake --build build\conda --parallel --target MODL MODL_Coclustering KhiopsNativeInterface KNITransfer _khiopsgetprocnumber
+cmake --build build\conda --parallel --target MODL MODL_Coclustering  KNITransfer _khiopsgetprocnumber
```
popescu-v: `KNITransfer` should be dropped and `KhiopsNativeInterface` should be kept.

popescu-v (changes requested):
See the pending comments.

packaging/conda/bld.bat
bruno-at-orange: done, sorry...

packaging/conda/meta.yaml
bruno-at-orange: you're right.
Now, it's working without the requirement

popescu-v (approved):
`:warning:` is not interpreted in the README.md file (perhaps use `[!WARNING]`?).
Otherwise, LGTM.

PR #1012, opened 2026-04-28
Support relative binary path arguments in kht_test

### LINKED ISSUE #1001: Take Relative Paths To the Binary Directory Into Account in LearningTestTool
Currently, LearningTestTool has an inconsistent behavior if a relative path to the binaries directory is provided and the said path is relative to the directory `kht_test.py` is launched into:
- `kht_test.py` runs correctly for the first test suite, or if a single test suite is provided;
- `kht_test.py` fails for the subsequent test suites, because it attempts to locate the binaries in a directory which is the concatenation of the testing suite directory of the previous test and the relative path to the binaries directory as provided.

Mitigations:
1. Automatically determine the absolute path to the binaries directory at the moment `kht_test.py` is launched.
2. Persist and `chdir` to the initial launch directory of `kht_test.py` at the end of each test, rather than to the testing suite directory of the said test.

Note: either 1. or 2. would address the issue, however:
- `1.` would be very localized and straightforward to implement, with minimal impact on the rest of the LearningTestTool code and behavior.
- `2.` would potentially be more side-effectful (to check), but would bring a more sensible behavior to each test:
   - at the beginning of the test: `chdir` to the testing directory
   - at the end of the test: `chdir` to the initial directory (where `kht_test.py` has been launched).
  This would allow each test suite to start in a consistent (and similar) state.

popescu-v:
closes #1001

### DIFF
--- a/test/LearningTestTool/py/kht_test.py
+++ b/test/LearningTestTool/py/kht_test.py
@@ -1111,6 +1111,12 @@ def main():
         test_dir_name,
     ) = utils.argument_parser_check_source_argument(parser, args.source)
 
+    # Pre-traitement de l'argument binaries: conversion en chemin absolu si pas d'alias
+    if args.binaries in [kht.ALIAS_D, kht.ALIAS_R]:
+        binaries_dir = args.binaries
+    else:
+        binaries_dir = os.path.abspath(args.binaries)
+
     # Verification des arguments optionnels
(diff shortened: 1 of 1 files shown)


marcboulle (approved):
LGTM

bruno-at-orange (approved):
LGTM

PR #1048, opened 2026-06-01
Improve load dll message

### LINKED ISSUE #1045: Message is missing when loading a driver with errors
When Khiops loads the GCS driver on Windows, there are currently dependency issues, but Khiops displays a misleading error message:
````cmd
warning : File driver libkhiopsdriver_file_gcs.dll : Library loading error (Ó│ÇµÁä╩ï)
````

### PR DESCRIPTION
**Pull request by bruno-at-orange**
- [Fix DLL loading error messages encoding and improve diagnostics](https://github.com/KhiopsML/khiops/commit/b10ce8c5cc6064e6426d8673f3c8f7383f2ff1af)
- [Improve DLL dependency search with LoadLibraryExW](https://github.com/KhiopsML/khiops/commit/21689ab3f33705b215cb943e94d16ca8d12cd938)

close #1045

### DIFF
--- a/src/Norm/base/Portability.cpp
+++ b/src/Norm/base/Portability.cpp
@@ -263,6 +263,10 @@ static void FormatErrorMessage(char* sErrorMessage)
 	DWORD dwErrorCode;
 	DWORD dwResult;
 	WCHAR wErrorMessage[SYSTEM_MESSAGE_LENGTH + 1];
+	int nRequiredBytes;
+	int nCopyLength;
+	int nTempBufferSize;
+	char* sTempBuffer;
 
 	// Capture du code d'erreur
 	dwErrorCode = GetLastError();
@@ -276,9 +280,7 @@ static void FormatErrorMessage(char* sErrorMessage)
 	// Utilise 0 pour la langue (langue systeme par defaut)
 	dwResult = FormatMessageW(FORMAT_MESSAGE_IGNORE_INSERTS | FORMAT_MESSAGE_FROM_SYSTEM, NULL, dwErrorCode,
 				  0, // Langue systeme par defaut
-				  wErrorMessage, SYSTEM_MESSAGE_LENGTH, NULL);
-
-	// Si FormatMessageW a echoue ou retourne vide
+				  wErrorMessage, SYSTEM_MESSAGE_LENGTH + 1, NULL);
 	if (dwResult == 0 || wErrorMessage[0] == L'\0')
 	{
 		snprintf(sErrorMessage, SYSTEM_MESSAGE_LENGTH, "System error %lu", dwErrorCode);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Norm/base/Portability.cpp [hunk 1]
```diff
@@ -285,6 +310,22 @@ static void FormatErrorMessage(char* sErrorMessage)
...
 	}
+
+	// Ajout d'un conseil specifique pour les erreurs courantes de chargement de DLL
+	// ERROR_MOD_NOT_FOUND (126) : module non trouve (souvent une dependance manquante)
+	// ERROR_BAD_EXE_FORMAT (193) : mauvais format exe (32/64 bits mismatch)
+	if (dwErrorCode == 126)
+	{
+		// L'erreur 126 signifie generalement qu'une dependance de la DLL est manquante
+		// Le message Windows est generique et ne precise pas quelle dependance manque
+		strncat(sErrorMessage, " This usually means a dependency DLL is missing.",
```
popescu-v: I would adopt an imperative form for the error message, e.g. "Check missing DLL dependencies". In this case, I would also add, if possible, a suggestion as to how to check which dependencies of the DLL are missing.

src/Norm/base/Portability.cpp
```diff
@@ -285,6 +310,22 @@ static void FormatErrorMessage(char* sErrorMessage)
...
+			SYSTEM_MESSAGE_LENGTH - strlen(sErrorMessage) - 1);
+	}
+	else if (dwErrorCode == 193)
+	{
+		strncat(sErrorMessage, " Check for 32/64-bit architecture mismatch.",
```
popescu-v: Could we add suggestion as to how to check this?

On the other hand, given that we create and ship the packages ourselves, we know they are 64-bit only. If we try to install them (Khiops or the drivers) on a 32-bit architecture, the install will fail before hitting this code anyway, right? So, I am not sure I see why this  `if` block is relevant to our context.

src/Norm/base/Portability.cpp
```diff
@@ -604,8 +604,29 @@ void* LoadSharedLibrary(const char* sLibraryPath, char* sErrorMessage)
 	nCurrentErrorMode = GetErrorMode();
 	SetErrorMode(SEM_FAILCRITICALERRORS);
 
-	// Chargement de la librairie
-	handle = (void*)LoadLibrary(wString);
+	// Chargement de la librairie avec LoadLibraryEx pour chercher les dependances dans le repertoire de la DLL
+	// LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR : cherche les dependances dans le repertoire de la DLL
+	// LOAD_LIBRARY_SEARCH_DEFAULT_DIRS : cherche dans les repertoires systeme et application
+	// Note : ces flags necessitent Windows Vista ou superieur
```
popescu-v, line 636: I think we support Windows 10 at the minimum, which is more recent than Vista. Hence, this note does not seem necessary to me: the install attempt on Vista should fail anyway, right?

src/Norm/base/Portability.cpp
```diff
@@ -604,8 +604,29 @@ void* LoadSharedLibrary(const char* sLibraryPath, char* sErrorMessage)
...
+	    (void*)LoadLibraryExW(wString, NULL, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
+
+	// Si echec, on sauvegarde l'erreur car elle est plus detaillee (ex: dependances manquantes)
+	DWORD dwFirstError = 0;
+	if (handle == NULL)
+	{
+		dwFirstError = GetLastError();
+
+		// Tentative avec LoadLibrary standard en fallback pour compatibilite
+		// (certains systemes peuvent ne pas supporter les flags de LoadLibraryEx)
```
popescu-v, line 647: Are we sure we need to support such systems (like, older than Windows 10)?

src/Norm/base/Portability.cpp
```diff
@@ -604,8 +604,29 @@ void* LoadSharedLibrary(const char* sLibraryPath, char* sErrorMessage)
...
+		handle = (void*)LoadLibraryW(wString);
+
+		// Si le fallback reussit, on garde handle
```
popescu-v: s/on garde handle/on garde le handle/

src/Norm/base/Portability.cpp
```diff
@@ -276,25 +277,39 @@ static void FormatErrorMessage(char* sErrorMessage)
...
+	if (nRequiredBytes <= SYSTEM_MESSAGE_LENGTH)
+	{
+		// La conversion tient dans le buffer de destination: conversion directe
+		WideCharToMultiByte(CP_UTF8, 0, wErrorMessage, -1, sErrorMessage, SYSTEM_MESSAGE_LENGTH, NULL, NULL);
+	}
+	else
+	{
+		// Conversion dans un buffer intermediaire puis troncature en respectant les sequences multi-octets UTF-8
+		// Chaque WCHAR produit au plus 4 octets UTF-8; le buffer couvre le pire cas
+		char sTempBuffer[4 * (SYSTEM_MESSAGE_LENGTH + 2)];
```
popescu-v: Why "`+ 2`"?

popescu-v (commented):
See comments and suggestions.

src/Norm/base/Portability.cpp
bruno-at-orange: done

src/Norm/base/Portability.cpp
bruno-at-orange: done

src/Norm/base/Portability.cpp
bruno-at-orange: block removed

src/Norm/base/Portability.cpp
bruno-at-orange: it's +1...

src/Norm/base/Portability.cpp
bruno-at-orange: No we support from windows 10. But I still prefer to keep the fallback (with windows I am paranoid)

src/Norm/base/Portability.cpp
bruno-at-orange: It's just a note

src/Norm/base/Portability.cpp
```diff
@@ -285,6 +330,18 @@ static void FormatErrorMessage(char* sErrorMessage)
 		else
 			break;
 	}
+
+	// Ajout d'un conseil specifique pour les erreurs courantes de chargement de DLL
+	// ERROR_MOD_NOT_FOUND (126) : module non trouve (souvent une dependance manquante)
+	// ERROR_BAD_EXE_FORMAT (193) : mauvais format exe (32/64 bits mismatch)
```
popescu-v: Drop this comment line, as error 193 handling has been removed.

src/Norm/base/Portability.cpp
popescu-v: Ok, fair enough.

popescu-v (changes requested):
Just a minor remaining comment.

src/Norm/base/Portability.cpp
bruno-at-orange: done

popescu-v (approved):
LGTM

PR #1053, opened 2026-06-04
Prepare release 11.0.1

bruno-at-orange:
- version 11.0.1-rc.1
- sign the binaries on pip/windows
- fix visualization desktop download for NSIS installer

### DIFF
--- a/.github/workflows/pack-nsis.yml
+++ b/.github/workflows/pack-nsis.yml
@@ -98,18 +98,18 @@ jobs:
           tar -zxf "$assets_tar_gz"
           cat assets/assets-info.env >> "$GITHUB_ENV"
           rm "$assets_tar_gz"
-      # We download the latest version of visualization tools:
-      # - They are backward compatible.
+      # We download the latest version of khiops-visualization-desktop:
+      # - It is backward compatible.
       # - The process is made easier by getting automatically the latest version
-      - name: Download latest releases of khiops-*visualization
+      - name: Download latest releases of khiops-visualization-desktop
         shell: bash
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           gh --repo https://github.com/KhiopsML/khiops-visualization-desktop release download --pattern *.exe -D visualization
           cd visualization
-          ls *
-          echo "KHIOPS_VIZ_FILENAME=$(ls khiops-visualization-desktop-Setup-*)" >> "$GITHUB_ENV"
+          echo "Downloaded files: $(ls *)"
+          echo "KHIOPS_VIZ_FILENAME=$(ls *)" >> "$GITHUB_ENV"
       - name: Download samples
         shell: bash
         env:
--- a/.github/workflows/pack-pip.yml
+++ b/.github/workflows/pack-pip.yml
@@ -8,6 +8,11 @@ on:
         type: choice
         default: None  # no publishing to any Package Index by default
         options: [None, testpypi, pypi]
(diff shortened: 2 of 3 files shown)


.github/workflows/pack-pip.yml
```diff
@@ -86,10 +93,69 @@ jobs:
...
+    needs: [build-wheel]
+    runs-on: windows-2025-vs2026
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/download-artifact@v8
+        with:
+          name: wheelhouse-windows-unsigned
+          path: wheelhouse
+      - name: Install Python
+        if: github.event_name == 'workflow_dispatch' && inputs.signature-activation
```
popescu-v: I'd group all steps that require this condition in a separate action that I would call `sign-wheel-windows` and I would use this action either in the `build` job (if possible), or in a separate eponymous job, rather than having this same condition on each individual step that requires it.

.github/workflows/pack-pip.yml
```diff
@@ -103,7 +169,7 @@ jobs:
     runs-on: ${{ matrix.setup.os }}
     container: ${{ matrix.setup.image }}
     name: Test wheel
-    needs: [build-wheel]
+    needs: [build-wheel, sign-wheel-windows]
```
popescu-v, line 175: Consider updating this according to the previous comment (if needed).

.github/workflows/pack-pip.yml
```diff
@@ -159,7 +225,7 @@ jobs:
             }}
   test-kni:
     name: Test KNI Python wheel
-    needs: [build-wheel]
+    needs: [build-wheel, sign-wheel-windows]
```
popescu-v, line 232: Consider updating this according to the previous comment (if needed).

.github/workflows/pack-nsis.yml
```diff
@@ -98,18 +98,18 @@ jobs:
...
           rm "$assets_tar_gz"
-      # We download the latest version of visualization tools:
-      # - They are backward compatible.
+      # We download the latest version of khiops-visualization-desktop:
+      # - It is backward compatible.
       # - The process is made easier by getting automatically the latest version
-      - name: Download latest releases of khiops-*visualization
+      - name: Download latest releases of khiops-visualization-desktop
         shell: bash
         env:
```
popescu-v, line 106: I would just use, on this step:
```
working-directory: visualization
```

In this case, the `-D visualization` snippet could be dropped from the `gh` command.

popescu-v (changes requested):
See the comments.

.github/workflows/pack-nsis.yml
bruno-at-orange: I prefer to keep it in the command line. It seems more explicit for me...

.github/workflows/pack-pip.yml
bruno-at-orange: I’m not sure that we need to create a composite action in a separate file for an action that will only be called in that one place. 
However, I added a variable SIGN_WINDOWS_WHEELS, which will be calculated once and for all

.github/workflows/pack-nsis.yml
popescu-v: Ok.

popescu-v (approved):

.github/workflows/pack-pip.yml
popescu-v: Ok for the variable.

PR #1055, opened 2026-06-09
Fix kht-test with 'check' parameters

bruno-at-orange:
fix the error:
running  `kht_test  ./LearningTest/ ckeck -f basic` results to 
 > Khiops binary /Users/bob/Dev/khiops/ckeck dir is not a valid directory


popescu-v (approved):
LGTM

PR #1067, opened 2026-06-19
Lookup KNITransfer in the same directory as khiops_env

### LINKED ISSUE #1066: KNI Tests are not Launched on Native Rocky Package Installations
In the `build_tool_exe_path` function in the `kht_test.py` Python file, the path to the `KNITransfer` executable is positioned with respect to the path of the `MODL` executable, as given by the `KHIOPS_PATH` environment variable, when `khiops_env` is used (that is, when this script is present in the binaries directory given to the `kht_test` command).

However, on native Rocky Linux installations, `khiops_env` is installed in `/usr/bin` (as `KNITransfer`, for that matter), but `KHIOPS_PATH` points to an OpenMPI-specific directory, viz. `/usr/lib64/openmpi/bin/khiops/`, where `MODL_openmpi` is situated.

This situation results in errors like: `error : Executable for tool name KNI not found` and the KNI tests are not launched.

The fix consists in setting the path to `KNITransfer` with respect to the directory `khiops_env` is situated in.

popescu-v:
Thusly, KNITransfer location is portable across OSes and environments, including on Rocky Linux native installations, where `dirname $KHIOPS_PATH != dirname $(which khiops_env)`.

### DIFF
--- a/test/LearningTestTool/py/kht_test.py
+++ b/test/LearningTestTool/py/kht_test.py
@@ -62,15 +62,23 @@ def build_tool_exe_path(tool_binaries_dir, tool_name, use_khiops_env):
             tool_exe_path = os.environ["KHIOPS_COCLUSTERING_PATH"]
         elif tool_name == kht.KNI:
             # cas particulier de KNITransfer, qui n'est pas positionne dans une variable d'environnement dans khiops_env,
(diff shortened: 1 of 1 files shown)


bruno-at-orange (approved):
LGTM

PR #1073, opened 2026-06-26
1072 kni learningtest tests fail on macos for the pip packages

### LINKED ISSUE #1072: KNI LearningTest tests fail on MacOS for the Pip Packages
`kht_test` fails to execute the KNI Tests on MacOs Pip installations of `khiops-core`, with errors like:
```python
dyld[...]: Library not loaded: @rpath/libKhiopsNativeInterface.11.dylib  Referenced from: <...> /path/to/khiops/testenv_khiops_11.0.1_py/bin/KNITransfer  Reason: tried: '/path/to/khiops/testenv_khiops_11.0.1_py/bin/../lib/libKhiopsNativeInterface.11.dylib' (no such file), '/path/to/khiops/testenv_khiops_11.0.1_py/bin/../lib/libKhiopsNativeInterface.11.dylib' (no such file)
```

Hence, KNITransfer looks for the KNI DLL in `env/lib`, whereas it is installed in `env/bin`.

This is because in `CMakeLists.txt`, `set_unix_khiops_options()` sets `INSTALL_RPATH` to `@loader_path/../lib/` for all pip targets. This makes `KNITransfer` (installed in `env/bin/`) look for `libKhiopsNativeInterface.XX.dylib` in `env/lib/`. However,  `install.cmake` installs it to `LIB_DIR = SKBUILD_SCRIPTS_DIR = env/bin/`.

On Linux, this works because `kht_test.py` sets `LD_LIBRARY_PATH` to `tool_exe_dir` before running `KNITransfer`, which consequently finds `libKhiopsNativeInterface.so.XX` in `env/bin/`, thus masking the same rpath bug. On macOS, `DYLD_LIBRARY_PATH` is not set in the same way, so the `@rpath` problem is exposed.

The fix consists in editing `src/Learning/KNITransfer/CMakeLists.txt` so that, on MacOS, the `KNITransfer` own dir is appended to its rpath so that is is found at runtime.

Another fix would be to update `LearningTestTool/py/kht_test.py` so that, in `evaluate_tool_on_test_dir`, `tool_exe_dir` is prepended to `DYLD_LIBRARY_PATH` on MacOS. But this fix only incidentally compensates for the root issue. The first fix in the CMake file addresses the root cause.

popescu-v:
closes #1072

### DIFF
--- a/.github/actions/run-standard-tests/action.yml
+++ b/.github/actions/run-standard-tests/action.yml
@@ -13,6 +13,18 @@ inputs:
   running-mode:
     description: Running mode (parallel or serial)
     required: true
+  test-khiops:
+    description: Test the MODL binary
+    required: false
+    default: 'true'
+  test-coclustering:
+    description: Test the MODL_Coclustering binary
+    required: false
+    default: 'true'
+  test-kni:
+    description: Test the KhiopsNativeInterface library via KNITransfer
+    required: false
+    default: 'true'
   config:
     description: Build configuration (debug or release)
     required: true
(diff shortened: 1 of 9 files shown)


.github/actions/run-standard-tests/action.yml
```diff
@@ -69,17 +81,27 @@ runs:
...
-          $PYTHON $TEST_PY test/LearningTest/TestKhiops/Standard "${{ inputs.binary-path }}" ${PARALLEL_ARG}
-          $PYTHON $TEST_PY test/LearningTest/TestCoclustering/Standard "${{ inputs.binary-path }}" ${PARALLEL_ARG}
-          if [[ "${{ inputs.running-mode }}" != "parallel" ]] ; then 
+          if [[ "${{ inputs.test-khiops }}" == "true" ]] ; then
+            $PYTHON $TEST_PY test/LearningTest/TestKhiops/Standard "${{ inputs.binary-path }}" ${PARALLEL_ARG}
+          fi
+          if [[ "${{ inputs.test-coclustering }}" == "true" ]] ; then
+            $PYTHON $TEST_PY test/LearningTest/TestCoclustering/Standard "${{ inputs.binary-path }}" ${PARALLEL_ARG}
+          fi
+          if [[ "${{ inputs.test-kni }}" == "true" && "${{ inputs.running-mode }}" != "parallel" ]] ; then
```
bruno-at-orange: I would not gate the KNI test with "${{ inputs.running-mode }}" != "parallel".

Explanation: we can run `kht_test LearningTest/TestKNI -p 4`, and KNITransfer is launched without MPI (because KNITransfer is not in kht.PARALLEL_TOOL_NAMES). This gives a smoother user experience: `kht_test LearningTest/ -p 4 `runs all tests in parallel for all tools in parallel except KNITransfer. We can follow the same approach here.

.github/workflows/pack-pip.yml
```diff
@@ -216,15 +216,27 @@ jobs:
...
         uses: ./.github/actions/run-standard-tests
         with:
           warning_as_error: true
           running-mode: parallel
           config: release
-          binary-path: ${{ steps.install_khiops_core.outputs.BIN_DIR }}
+          binary-path: ${{ steps.install_khiops_core_and_kni.outputs.BIN_DIR }}
+          testing-context-description: pip-test-${{ matrix.setup.os }}-${{ matrix.setup.image
+            }}
+      - name: Run standard tests on the KNI package
```
bruno-at-orange: We could remove this extra call to `run-standard-tests` if the KNI test is also run in parallel mode.

bruno-at-orange (changes requested):
One question on the action `run-standard-tests`

.github/actions/run-standard-tests/action.yml
popescu-v: OK for me. It was gated before, so I had kept this, but OK for removing it.

.github/workflows/pack-pip.yml
popescu-v: OK, but the tests would go in the same step then. Ok with this @bruno-at-orange ?

.github/actions/run-standard-tests/action.yml
popescu-v: Done: parallel gate removed.

.github/workflows/pack-pip.yml
popescu-v: Done: conflated the two tests into a single steps that runs Core and KNI tests.

bruno-at-orange (approved):
LGTM

PR #1083, opened 2026-07-03
Add rebase=never to renovate config

bruno-at-orange:
Renovate will stop to rebase all the time.
In the other hand we have to check the box "rebase and test" before merging


popescu-v (approved):
LGTM. Nevertheless, I would add an entry to the wiki explaining the maintenance of these renovate-originated branches.

PR #1101, opened 2026-07-27
Fix khiops installation for tests by pinning the version

sgouache:
Tests fail due to wrong version of khiops being installed. This is due to the behavior of pip that prefers repository packages over local files.
This is more a fragile workaround than a fix: we explicitly pin the versions to be installed by using the locally built version numbers. If such versions do not exist in pypi registry the local versions will be installed.


marcboulle (approved):
LGTM

PR #1132, opened 2026-08-26
1109 use   no index with pip install   find links

### LINKED ISSUE #1109: Use `--no-index` with  `pip install --find-links`
In the PR https://github.com/KhiopsML/khiops/pull/1101 It would have been better to use `--no-index`

bruno-at-orange:
fix: install wheels from the local wheelhouse
Install the wheel with 2 steps:
- install the wheel itself only from the local wheelhouse with --find-links.
   The flag --no-index is used to ignore the packages index (and then from installing old wheels).
- install the runtime dependencies (not found in the previous install without the packages index)

Out of scope:
- In pyproject-knitransfer.toml: extract the version of kni from KWKhiopsVersion.h

### DIFF
--- a/.github/workflows/pack-pip.yml
+++ b/.github/workflows/pack-pip.yml
@@ -69,6 +69,8 @@ jobs:
           else
             source "$KHIOPS_INSTALL_DIR/bin/activate"
           fi
+          # --only-deps requires pip>=26.2
+          python -m pip install --upgrade "pip>=26.2"
           # Install the wheel itself only from the local wheelhouse, then resolve its
           # runtime dependencies (e.g. openmpi) from PyPI
           python -m pip install --find-links=wheelhouse --no-index --no-deps khiops-core
@@ -219,6 +221,8 @@ jobs:
             source "$KHIOPS_INSTALL_DIR/bin/activate"
             echo "BIN_DIR=$KHIOPS_INSTALL_DIR/bin" >> "$GITHUB_OUTPUT"
           fi
+          # --only-deps requires pip>=26.2
+          $PYTHON -m pip install --upgrade "pip>=26.2"
           # Install the wheels themselves only from the local wheelhouse, then resolve their
           # runtime dependencies (e.g. openmpi) from PyPI
           $PYTHON -m pip install --find-links=wheelhouse --no-index --no-deps khiops-core khiops-kni khiops-knitransfer
@@ -266,6 +270,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install khiops-kni wheel
         run: |
+          # --only-deps requires pip>=26.2
+          python -m pip install --upgrade "pip>=26.2"
           # Install the wheel itself only from the local wheelhouse, then resolve its
           # runtime dependencies from PyPI
           python -m pip install --find-links=wheelhouse --no-index --no-deps khiops-kni
--- a/.github/workflows/pack-pip.yml
+++ b/.github/workflows/pack-pip.yml
@@ -201,7 +201,7 @@ jobs:
               apt-get install -y python3 > /dev/null
             else
               yum check-update || true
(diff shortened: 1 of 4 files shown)


.github/workflows/pack-pip.yml
```diff
@@ -69,6 +69,9 @@ jobs:
           else
             source "$KHIOPS_INSTALL_DIR/bin/activate"
           fi
+          # Install the wheel itself only from the local wheelhouse, then resolve its
+          # runtime dependencies (e.g. openmpi) from PyPI
+          python -m pip install --find-links=wheelhouse --no-index --no-deps khiops-core
           python -m pip install --find-links=wheelhouse khiops-core
```
popescu-v, line 75: IMHO this approach incurs the risk of (re) installing khiops-core from PyPI if, for any reason, local khiops-core version is deemed "older" than the PyPI version. A more direct and IMHO safer approach has been adopted in khiops-python to this end, along the lines:
- extract dependencies from pyproject.toml
- install these dependencies.
See https://github.com/KhiopsML/khiops-python/blob/015e844d18aeefc5d309016095b743555d35c6bf/.github/workflows/pip.yml#L154 for the details (also, the Python scripts for extracting the dependencies is at https://github.com/KhiopsML/khiops-python/blob/main/scripts/extract_dependencies_from_pyproject_toml.py).

Perhaps it can be simplified / adapted to the current context and used as in the khiops-python CI. 

More specifically, I would consider:

- extracting khiops-core dependencies from the pyproject.toml and installing them from PyPI
- afterwards, installing khiops-core from the locally-produced wheel.

Thus, IMHO we make sure to control precisely which packages are installed from where.

.github/workflows/pack-pip.yml
```diff
@@ -216,17 +219,10 @@ jobs:
...
-          KNI_VERSION=$(ls wheelhouse/khiops_kni-*.whl | head -1 | xargs basename | cut -d- -f2)
-          KNITRANSFER_VERSION=$(ls wheelhouse/khiops_knitransfer-*.whl | head -1 | xargs basename | cut -d- -f2)
-          $PYTHON -m pip install \
-            --find-links=wheelhouse \
-            "khiops-core==$CORE_VERSION" \
-            "khiops-kni==$KNI_VERSION" \
-            "khiops-knitransfer==$KNITRANSFER_VERSION"
+          # Install the wheels themselves only from the local wheelhouse, then resolve their
+          # runtime dependencies (e.g. openmpi) from PyPI
+          $PYTHON -m pip install --find-links=wheelhouse --no-index --no-deps khiops-core khiops-kni khiops-knitransfer
```
popescu-v, line 224: Same comment as above.

.github/workflows/pack-pip.yml
```diff
@@ -269,7 +265,12 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install khiops-kni wheel
-        run: python -m pip install --find-links=wheelhouse khiops-kni
+        run: |
+          # Install the wheel itself only from the local wheelhouse, then resolve its
+          # runtime dependencies from PyPI
+          python -m pip install --find-links=wheelhouse --no-index --no-deps khiops-kni
```
popescu-v, line 271: Same comment as above.

popescu-v (changes requested):
Basically, 1 comment repeated in several places.

popescu-v (changes requested):
One point, in several places (see the comment and its referrals).

.github/workflows/pack-pip.yml
bruno-at-orange: I found a better solution:  the flag `--only-deps`of `pip install. It's available since pip 26.2, as its name suggests it installs only dependencies. I have tested it with a package versioned 8.0.0

popescu-v (approved):
LGTM

